PR #18, opened 2017-07-22
Initial UI and bootstrapping for infrastructure

romanofski:
Mostly bootstrapping commits for:

* CI (travis)
* UI architecture using two modes: mail index and mail view


romanofski:
I'm merging this now... just because at this point it's to early to consider whatever we have serious, but it might help others to get a glimpse ...

PR #32, opened 2017-08-10
HACKING: avoid infix optic functions, require explicit import lists

frasertweedale:
add some stuff to HACKING.

The import list thing is hopefully uncontroversial.

The optic thing is a strong preference I have (mainly because I can never remember
what they are, but the "word" ones are much more human friendly and therefore readable (IMO).

If you agree, no need to wholesale refactor.  Drive-bys are fine.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -13,6 +13,11 @@ communication is the key here rather than hacking away.
 
 - optics are your friend
 
+- use `view`, `review`, `set`, `over` and so on instead of the
+  infix optic functions from lens
+
+- use explicit import list or qualified imports
+
 - no perf optimisations without measurements (preferably in the commit
   message)
 


romanofski:
All fine by me :-)

frasertweedale:
@romanofski do you reckon we could rebase PRs instead of merge commits (in future)?
Keeps the history a bit cleaner and easier to follow IMO.

romanofski:
Sounds good. Since we're on the topic: I've been using [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) for most of the imports and function formatting. I think the only thing which irks me is the fact that some stylish changes reformat the entire import block depending on the width of the imports. So I'm not sure if it will make hard to follow changes in the future if they're just clobbered by various "styling" changes. What are your thoughts on this?

frasertweedale:
@romanofski I am against aligning import lists.  If a new (or removed) import
changes the width, everything is then misaligned.

i.e. **do not want** the following:

```
import qualified Foo     (a, b)
import           Bar.Baz (c, d)
```

, because when we need to add `import Quuz.Xyzzy` everything gets misaligned.
Like you said, if you use a code reformatter it results in spurious and noisy diff.
Do not want.

I am ambivalent about aligning for present/absense of `qualified` (as 
above), but lean against *requiring* it, for similar reasons.

romanofski:
Alright good to know... Then I ditch the formatter :-)

frasertweedale:
👍

Issue #50, opened 2017-08-30
`confEditor` config is unused


frasertweedale:
We are not using this config.  When we invoke the editor to inspect `EDITOR` env var,
and default to `vi` explicitly at that point.

This may be another case of "default config is an `IO a`"...

The default action SHOULD use whatever is in `EDITOR` if set, otherwise use vi.

Thinking more about this it is a nice idea to have `IO FilePath` for this config too.
Use case: a user wants a single purebred config to work on different hosts, and the
config uses aspects of the environment to work out what to invoke.  For example if
we end up supporting Windows, the action could inspect the platform / environment / 
see what's installed and select the editor accordingly.

romanofski:
Hah yes, we totally ignore that. 

I'm not really sure I understand why we need an `IO FilePath` for this, since it seems to me that it is something easily configurable without any IO interaction.

Even then, using just `IO FilePath` might actually not be enough tho. Since it's not only the editor you want, you might also want to pass arguments to the editor. E.g.

  vim -c 'set tw=72' -f +8

At the moment purebred just takes what's in the EDITOR variable and executes it in a shell which includes any arguments EDITOR is configured with.

frasertweedale:
On Thu, Aug 31, 2017 at 12:49:16AM +0000, Roman Joost wrote:
> Hah yes, we totally ignore that. 
> 
> I'm not really sure I understand why we need an `IO FilePath` for
> this, since it seems to me that it is something easily
> configurable without any IO interaction.
> 
Don't you want the default behaviour to inspect `EDITOR` envvar?
Reading the environment is an IO computation:

    lookupEnv :: String -> IO (Maybe String)

> Even then, using just `IO FilePath` might actually not be enough
> tho. Since it's not only the editor you want, you might also want
> to pass arguments to the editor. E.g.
> 
>   vim -c 'set tw=72' -f +8
> 
Sure, then instead of `IO FilePath` we make it an `IO (FilePath,
[String])`, or something like that.

romanofski:
Ah I understand now. Instead of performing the environment variable lookup somewhere in the code (deep down in UI.GatherHeaders.Main), rather make it the default in the config and let the user overwrite it with his config, right?

Then maybe leave it as `IO String` ? It seems for this kind of thing, using a shell is fine (don't need that much control over the editor process).

romanofski:
This is merged.

PR #52, opened 2017-08-31
Honor the editor configured in the config

romanofski:
Always pick the editor configured in notmuch if EDITOR is unset.

Issue: #50

### DIFF
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -4,8 +4,9 @@ module Config.Main where
 import qualified Brick.AttrMap as A
 import Brick.Util (fg, on)
 import qualified Brick.Widgets.Edit as E
-import qualified Brick.Widgets.List as L
 import qualified Graphics.Vty as V
+import System.Environment (lookupEnv)
+import Data.Maybe (fromMaybe)
 import UI.Mail.Main (headerKeyAttr, headerValueAttr)
 import UI.ComposeEditor.Keybindings (composeEditorKeybindings)
 import UI.Index.Main
@@ -45,7 +46,7 @@ defaultConfig =
       , _nmDatabase = getDatabasePath
       , _nmNewTag = "unread"
       }
-    , _confEditor = "vi"
+    , _confEditor = fromMaybe "vi" <$> lookupEnv "EDITOR"
     , _confMailView = MailViewSettings
       { _mvIndexRows = 10
       , _mvPreferedContentType = "text/plain"
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -46,6 +46,7 @@ import Graphics.Vty.Input.Events (Event(..), Key(..), Modifier(..))
 import Brick.Main (halt, continue, defaultMain)
 import Control.Lens.Lens ((&))
 import Control.Lens.Setter (over, set)
+import Control.Lens.Getter (view)
 
 data AppConfig = AppConfig
     { databaseFilepath :: Maybe String
@@ -99,7 +100,12 @@ launch cfg = do
     void $ defaultMain (theApp s) s
 
 processConfig :: UserConfiguration -> IO InternalConfiguration
-processConfig = (confNotmuch . nmDatabase) id
+processConfig cfg = do
+    fp <- view (confNotmuch . nmDatabase) cfg
+    ed <- view confEditor cfg
+    pure $ cfg
+      & set (confNotmuch . nmDatabase) fp
+      & set confEditor ed
 
 -- | Recompile the config file if it has changed based on the modification timestamp
 -- Node: Mostly a XMonad.Main.hs rip-off.
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -116,34 +116,34 @@ nmDatabase f (NotmuchSettings a b c) = fmap (\b' -> NotmuchSettings a b' c) (f b
 nmNewTag :: Getter (NotmuchSettings a) T.Text
 nmNewTag = to (\(NotmuchSettings _ _ c) -> c)
 
-data Configuration a = Configuration
+data Configuration a b = Configuration
     { _confColorMap :: Brick.AttrMap
     , _confNotmuch :: NotmuchSettings a
-    , _confEditor :: T.Text
+    , _confEditor :: b
     , _confMailView :: MailViewSettings
     , _confIndexView :: IndexViewSettings
     , _confComposeView :: ComposeViewSettings
     }
 
-type UserConfiguration = Configuration (IO FilePath)
-type InternalConfiguration = Configuration FilePath
+type UserConfiguration = Configuration (IO FilePath) (IO String)
+type InternalConfiguration = Configuration FilePath String
 
-confColorMap :: Getter (Configuration a) Brick.AttrMap
+confColorMap :: Getter (Configuration a b) Brick.AttrMap
 confColorMap = to (\(Configuration a _ _ _ _ _) -> a)
 
-confEditor :: Lens' (Configuration a) T.Text
+confEditor :: Lens (Configuration a b) (Configuration a b') b b'
 confEditor f (Configuration a b c d e g) = fmap (\c' -> Configuration a b c' d e g) (f c)
 
-confNotmuch :: Lens (Configuration a) (Configuration b) (NotmuchSettings a) (NotmuchSettings b)
+confNotmuch :: Lens (Configuration a c) (Configuration b c) (NotmuchSettings a) (NotmuchSettings b)
 confNotmuch f (Configuration a b c d e g) = fmap (\b' -> Configuration a b' c d e g) (f b)
 
-confMailView :: Lens' (Configuration a) MailViewSettings
+confMailView :: Lens' (Configuration a b) MailViewSettings
 confMailView f (Configuration a b c d e g) = fmap (\d' -> Configuration a b c d' e g) (f d)
 
-confIndexView :: Lens' (Configuration a) IndexViewSettings
+confIndexView :: Lens' (Configuration a b) IndexViewSettings
 confIndexView f (Configuration a b c d e g) = fmap (\e' -> Configuration a b c d e' g) (f e)
 
-confComposeView :: Getter (Configuration a) ComposeViewSettings
+confComposeView :: Getter (Configuration a b) ComposeViewSettings
 confComposeView = to (\(Configuration _ _ _ _ _ h) -> h)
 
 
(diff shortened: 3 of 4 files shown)


frasertweedale:
On Thu, Aug 31, 2017 at 01:05:07AM +0000, Roman Joost wrote:
> @romanofski requested your review on: purebred-mua/purebred#52 Honor the editor configured in the config.
> 
Thx, I'll review tonight or maybe tomorrow.

romanofski:
Will do!

romanofski:
Ugh... right against the wall.

I've got:

```haskell
data NotmuchSettings m = NotmuchSettings
    { _nmSearch :: T.Text
    , _nmDatabase :: m FilePath
    , _nmNewTag :: T.Text
    }

nmDatabase :: Lens' (NotmuchSettings m) (m FilePath)
nmDatabase = lens _nmDatabase (\nms fp -> nms { _nmDatabase = fp })

type UserConfiguration = Configuration IO
type InternalConfiguration = Configuration Identity
```
So far so good. But going from UserConfiguration to InternalConfiguration is where I'm lost. How the hell do I go from IO -> Identity?

frasertweedale:
@romanofski follow the types.  What can you do with the type of a lens (including a
composed lens)?

```haskell
type Lens s t a b = (a -> f b) -> s -> f t
```

Given the target type:

```haskell
processConfig :: Configuration IO -> Configuration Identity
```

How can you use a lens to reach the goal?

Let me know if you are still stuck after a while :)

romanofski:
I've found a way to work my brain around a solution. Not sure if it's convoluted, but type checks:

```Haskell
processConfig :: UserConfiguration -> IO InternalConfiguration
processConfig cfg = do
  fp <- view (confNotmuch . nmDatabase) cfg
  pure $ set (confNotmuch . nmDatabase) (pure fp) cfg
```

frasertweedale:
@romanofski well done.  Isn't type tetris fun!

Type golf is even better :)

```haskell
processConfig :: UserConfiguration -> IO InternalConfiguration
processConfig = (confNotmuch . nmDatabase) (fmap pure)
```

I often find it helps to think of an optic (lenses in particular) like `traverse` - gives you a way of
propagating an effect at the focus of the lens, out to the outside of the structure.  In the previous code we had:

```haskell
processConfig :: Configuration (IO FilePath) -> IO (Configuration FilePath)
processConfig = (confNotmuch . nmDatabase) id
```

So above, `id` give us `IO FilePath` - the `IO` gets pushed to the "outside" and we keep `FilePath` as the type at the focus.

Now in the new code:
```haskell
processConfig :: Configuration IO -> IO (Configuration Identity)
processConfig = (confNotmuch . nmDatabase) (fmap pure)
```
we have `fmap pure :: IO FilePath -> IO (Identity FilePath)`.  Again the `IO` gets propagated out,
and we keep `Identity FilePath` as the type at the focus, which is precisely what we need.

HTH!

romanofski:
Cheers for the explanation. Had to quickly push my changes, so still working on it :)

src/Config/Main.hs
```diff
@@ -45,7 +46,7 @@ defaultConfig =
       , _nmDatabase = getDatabasePath
       , _nmNewTag = "unread"
       }
-    , _confEditor = "vi"
+    , _confEditor = lookupEnv "EDITOR" >>= pure . fromMaybe "vi"
```
frasertweedale: This can be `fromMaybe "vi" <$> lookupEnv "EDITOR"`

frasertweedale:
For reading the `Identity`-wrapped fields, have a look at `Control.Lens.Wrapped`.
Lets you get at the thing inside a "wrapping" type, which includes `Identity`.

```
λ> import Data.Functor.Identity (Identity(..))
λ> import Control.Lens (view, _Wrapped)
λ> view _Wrapped (Identity "hello")
"hello"
```

Then instead of importing `runIdentity` and applying it to the result of a `view`, you
can compose the `_Wrapped` iso into the optic, e.g.:

```haskell
  let editor = view (asConfig . confEditor . _Wrapped) s 
```

romanofski:
Ah cheers for the hint.

I'm a bit puzzled that I couldn't repeat my luck I had before by "dumbing" it down a bit. So again I have:

```Haskell
processConfig :: UserConfiguration -> IO InternalConfiguration
processConfig cfg = do
    fp <- view (confNotmuch . nmDatabase) cfg
    editor <- view confEditor cfg
    let cfg' =
            set (confNotmuch . nmDatabase) (pure fp) cfg &
            set confEditor (pure editor)
    pure cfg'
```
but I get the error:

```
    • Couldn't match type ‘IO’ with ‘Data.Functor.Identity.Identity’
      Expected type: IO InternalConfiguration
        Actual type: IO (Configuration IO)
    • In a stmt of a 'do' block: pure cfg'
      In the expression:
        do { fp <- view (confNotmuch . nmDatabase) cfg;
             editor <- view confEditor cfg;
             let cfg'
                   = set (confNotmuch . nmDatabase) (pure fp) cfg
                     & set confEditor (pure editor);
             pure cfg' }
      In an equation for ‘processConfig’:
          processConfig cfg
            = do { fp <- view (confNotmuch . nmDatabase) cfg;
                   editor <- view confEditor cfg;
                   let cfg' = ...;
                   .... }
```
which I'm surprised about. I'm "pulling" the String out of the IO computation so my reasoning is, that I should be able to "set" it back into the config, but it seems the compiler still assuming there is still an IO type there?

frasertweedale:
I think you want something like:

```haskell
processConfig :: UserConfiguration -> IO InternalConfiguration
processConfig cfg = do
    fp <- view (confNotmuch . nmDatabase) cfg
    editor <- view confEditor cfg
    pure $ cfg
      & set (confNotmuch . nmDatabase) (pure fp)
      & set confEditor (pure editor)
```

romanofski:
Hm.. nope. Computer says: "no".

frasertweedale:
@romanofski uh, so it turns out we need just as many type variables after all, otherwise
we get type mismatches during the "transitional" phase when processing multiple fields.
I'm undecided which approach is worst... the old way we were doing it or the new way.  But I am leaning towards the OLD way now.

FWIW here is (an example of) what you'd have to do under the current paradigm:
```haskell
data Config' m m1 = Config (m String) (m1 String)
type Config m = Config' m m
type UserConfig = Config IO
type InternalConfig = Config Identity

db :: Lens (Config' m m1) (Config' m' m1) (m String) (m' String)
db f (Config a b) = fmap (\a' -> Config a' b) (f a)

ed :: Lens (Config' m m1) (Config' m m1') (m1 String) (m1' String)
ed f (Config a b) = fmap (\b' -> Config a b') (f b)

processConfig :: UserConfig -> IO InternalConfig
processConfig = db (fmap pure) >=> ed (fmap pure)
```

I really think I prefer the old way, which would be:
```
data Config a b = Config a b
type UserConfig = Config (IO String) (IO String)
type InternalConfig = Config String String

db :: Lens (Config a b) (Config a' b) a a'
db f (Config a b) = fmap (\a' -> Config a' b) (f a)

ed :: Lens (Config a b) (Config a b') b b'
ed f (Config a b) = fmap (\b' -> Config a b') (f b)

processConfig :: UserConfig -> IO InternalConfig
processConfig = db id >=> ed id
```

Mainly because if we need all the typevars anyway, I'd rather have them correspond
to inhabitable type (just seems "easier to understand" that way).  It also allows to
vary types between UserConfig and InternalConfig more, and we can do away with
the `_Wrapped` business that was necessary when abstracting over `(* -> *)` and using
`Identity` for the internal config type.

So... sorry for taking us down the garden path on this one, but I think we should change back :)

romanofski:
Yeah I'd also opt for an easier to understand approach. At least this doesn't seem to warrant the complexity. No worries, I'll change it back. Not today anymore tho.

romanofski:
Still can't make the type system happy. Would you mind to check my `confNotmuch` lens if it might create the error I run into in `Purebred.hs:105:9`?

```
    • Couldn't match type ‘[Char]’ with ‘IO FilePath’
        arising from a functional dependency between:
          constraint ‘mtl-2.2.1:Control.Monad.Reader.Class.MonadReader
                        (Configuration (IO (IO FilePath)) (IO String))
                        ((->) UserConfiguration)’
            arising from a use of ‘view’
          instance ‘mtl-2.2.1:Control.Monad.Reader.Class.MonadReader
                      r ((->) r)’
            at <no location info>
    • In a stmt of a 'do' block:
        fp <- view (confNotmuch . nmDatabase) cfg
      In the expression:
        do { ed <- view confEditor cfg;
             fp <- view (confNotmuch . nmDatabase) cfg;
             pure
             $ cfg & set (confNotmuch . nmDatabase) fp & set confEditor ed }
      In an equation for ‘processConfig’:
          processConfig cfg
            = do { ed <- view confEditor cfg;
                   fp <- view (confNotmuch . nmDatabase) cfg;
                   pure
                   $ cfg & set (confNotmuch . nmDatabase) fp & set confEditor ed }
```

Is it me who's not seeing it or maybe my botched lens?

frasertweedale:
looks like somehow you got an `IO (IO Filepath)` in there.  Feel free to commit WIP and I'll take a look at it.

romanofski:
Latest WIP is already up, but let me give it another go to figure out what's wrong. Maybe I find the issue ...

romanofski:
Oohh I've figured it out. W00t. Will push asap...

romanofski:
Just to write it down for myself: I've made a mistake with the lens and it's type signature. While it was type checking, I basically "mapped" it to the wrong types. This ended either up picking the "editor" (IO String) or a double IO Monad. Splitting up the first `view (confNotmuch . nmDatabase) cfg` into smaller pieces helped to figure out what was wrong.

src/Config/Main.hs
frasertweedale: ... this simplification is still applicable ;)

src/Purebred.hs
```diff
@@ -99,7 +100,10 @@ launch cfg = do
     void $ defaultMain (theApp s) s
 
 processConfig :: UserConfiguration -> IO InternalConfiguration
-processConfig = (confNotmuch . nmDatabase) id
+processConfig cfg = do
+    fp <- view (confNotmuch . nmDatabase) cfg
+    ed <- view confEditor cfg
+    pure $ cfg & set (confNotmuch . nmDatabase) (fp) & set confEditor (ed)
```
frasertweedale: would be good to break this across multiple lines & remove extraneous parens, e.g.
```haskell
pure $ cfg
  & set (confNotmuch . nmDatabase) fp
  & set confEditor ed
```

romanofski: Ah dang! Yes of course. Will do.

Issue #51, opened 2017-08-31
Search which yields a large results takes too long


romanofski:
A search which yields a large result takes quite a long time. I've profiled a bit and here are the stats:

Mail indexed in notmuch: 170297
notmuch search: *

* neomutt: 18s
* purebred: 32s
* alot: 1s (not even)

[purebred.txt](https://github.com/purebred-mua/purebred/files/1265415/purebred.txt)

frasertweedale:
@romanofski awesome, thanks for the cost centres.

Do the programs definitely do the same thing?  For example, you are
not just loading every message but reading the tags of every
message.

Here's a few quick takeaways:

- detachPtr is surprisingly expensive (> 10%) but it can't be
  improved.

- message_get_filename is individually responsible for ~65% of the
  allocation of the entire program.  This is due to reading the
  filename into a `String` i.e. a cons list of `Char` (and the
  filenames tend to be longish).

  Recommendation: defer reading the filename until we actually need
  it.  Alternatively: change messageFilename to return a ByteString
  instead (still allocs, but should be way cheaper).

- Copying tags is an interesting case - we will be copying the same
  string over and over again.  Interning them could save a lot of
  copies (maybe notmuch itself interns them, though if it does we
  probably can't rely on this).  In any case, the main overhead here
  currently seems to be time spent iterating the notmuch_tags_t
  (~28% of program runtime), not alloc (~7%).

  Of that 28%, ~9% is notmuch_tags_valid and comparing the result to
  see if there are more items, ~5% is (notmuch_tags_get + copying
  the tag value + cons cell alloc), ~5% is notmuch_tags_move_to_next
  (advancing the iterator).  So ~15%..20% of the 28% is absolutely
  irreducible.  Similar ratios will apply for other iterators
  (messages, threads, etc) but there is a smaller number of those so
  it has a smaller impact overall.

- messageHeader is kind of expensive (~14% of runtime).  I suspect
  this is due to reading one of the headers that is not indexed
  ("To:").  If you avoid reading the To: header, how much does the
  runtime improve?

- You are reading the tags twice!  That's 28% of the execution time
  EACH.  The offending code is in `messageToMail` - you call
  tagsToText, then also call `isNewMail` which also gets the tags.
  Get the tag list ONCE and use it twice :)  Poof!  28% of the
  program runtime will vanish :)

romanofski:
[...]

PR #54, opened 2017-08-31
Don't parse tags twice

romanofski:
This has quite an impact when showing a huge chunk of mail (28% of
execution time per message).

Issue: #51

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -36,20 +36,19 @@ messageToMail
     => T.Text
     -> Message
     -> IO NotmuchMail
-messageToMail ignoredTag m =
+messageToMail ignoredTag m = do
+    tgs <- tags m
     NotmuchMail <$>
-    (decodeUtf8 . fromMaybe "" <$> messageHeader "Subject" m) <*>
-    (decodeUtf8 . fromMaybe "" <$> messageHeader "To" m) <*>
-    (decodeUtf8 . fromMaybe "" <$> messageHeader "From" m) <*>
-    messageFilename m <*>
-    messageDate m <*>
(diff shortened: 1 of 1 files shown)


src/Storage/Notmuch.hs
```diff
@@ -36,20 +36,19 @@ messageToMail
...
+      messageFilename m <*>
+      messageDate m <*>
+      (tagsToText tgs ignoredTag) <*>
+      isNewMail tgs ignoredTag
 
-tagsToText :: HasTags a => a -> T.Text -> IO [T.Text]
-tagsToText m ignored = do
-  t <- tags m
-  pure $ filter (/= ignored) $ decodeUtf8 <$> t
+tagsToText :: [Tag] -> T.Text -> IO [T.Text]
```
frasertweedale: This should be a pure function.  If the caller want to lift it into `IO` let
them apply `pure` at the call site.

src/Storage/Notmuch.hs
```diff
@@ -61,7 +60,5 @@ getFromNotmuchConfig key = do
   stdout <- readProcess cmd args []
   pure $ filter (/= '\n') stdout
 
-isNewMail :: HasTags Message => Message -> T.Text -> IO Bool
-isNewMail m newTag = do
-  t <- tags m
-  pure $ newTag `elem` (decodeUtf8 <$> t)
+isNewMail :: [Tag] -> T.Text -> IO Bool
```
frasertweedale: Same here; this should be pure, and furthermore should be abstracted over `Foldable t` i.e.

```haskell
isNewMail :: Foldable t => t Tag -> T.Text -> Bool
```

romanofski: Yeah good point actually!

PR #56, opened 2017-09-03
Automatic mark mail as read

romanofski:


src/Storage/Notmuch.hs
```diff
@@ -31,9 +35,50 @@ getMessages s settings = do
...
+setNotmuchMailTags
+    :: FilePath
+    -> NotmuchMail
+    -> IO (Either String NotmuchMail)
+setNotmuchMailTags dbpath m =
+    case mailTagsToNotmuchTags m of
+        Nothing -> pure $ Left ("Tags are corrupt")
+        Just nmtags ->
+            bracket
+                (databaseOpen dbpath >>= either (error . show) pure)
```
romanofski: I've fiddled with the `error` here. This would obviously crash, but the Type expected in IO is always (Database RW), but I'm puzzled, since I thought the first function handles the Left value, which is `String`. What I want is to give an error back, rather than... well just crashing.

romanofski:
Travis died as well with the change. Will have to investigate. Wouldn't be surprised if it has something to do with the old, but stable version of Ubuntu *sigh*. I'm also thinking of adding one additional acceptance test to check if toggling works. Not sure if it will run in Travis tho, since I've noticed that I dealt with different escape sequences to assert e.g. text is bold or not.

src/Storage/Notmuch.hs
frasertweedale: We're going to need to start using MonadError monad transformer to be able to better manage *IO-with-error* scenarios.  For now leave it as-is and I'll address in a later commit (just add a FIXME comment)

romanofski: I've moved it in a backlog item... when we have the time.

romanofski:
Hm.. I've got trouble building `hs-notmuch` now (see travis error). I actually ran into the same error locally as well, but I was unable to reproduce it 100%. In case you know what's going on on top of year head please share :))

frasertweedale:
@romanofski there are breaking API changes in notmuch v0.25 :(  Might take a little while to sort out.

romanofski:
@frasertweedale I'd be happy to add in a .travis and stack.yml file. stack just because most of the infrastructure is now build around it, including ready-made files for testing against several platforms/GHC.

frasertweedale:
@romanofski IMO the "correct" approach for *hs-notmuch* is to determine which versions of notmuch we want to support, and have CI that checks out every single release, builds libnotmuch, then builds hs-notmuch against it.  But that's a task for another day.

frasertweedale:
@romanofski well I didn't quite get it done today.  But the changes are well on the way.  Hopefully Wednseday.

frasertweedale:
@romanofski new *hs-notmuch* commit which should hopefully fix CI issues, add support for notmuch-0.25, and introduces the monad transformer for dealing with errors. Lots of API changes, you will have fun chasing this.

purebred-mua/hs-notmuch@5bc75a6344163899d09153bb9e08bc620ee343ee

The most important function you'll need to perform the refactor/rebase is https://hackage.haskell.org/package/mtl-2.2.1/docs/Control-Monad-Except.html#v:runExceptT.

romanofski:
Does it build?

romanofski:
Meh now I realize that I've squashed my changes against latest hs-notmuch just with the parent. Should have perhaps fiddled it into the separate patches :/ Could probably do that with a few items I still want to "clean up"

romanofski:
Alright doesn't build... perhaps we should yank hs-notmuch into travis first ...

romanofski:
I think I'm getting there to where I want to have this and you can start having a look. I think the intermittend test failures are caused by unstable search results, since the maildir I'm using exposes two mails with different content. That's why I created purebred-mua/hs-notmuch#15

Anyway, have a look and tell me what you think.

src/Storage/Notmuch.hs
```diff
@@ -1,39 +1,86 @@
...
+getMessages :: T.Text
+            -> NotmuchSettings FilePath
+            -> IO (Either String (Vec.Vector NotmuchMail))
+getMessages s settings = runExceptT
+  (do msgs <- databaseOpenReadOnly (view nmDatabase settings)
+              >>= flip query (FreeForm $ T.unpack s)
+              >>= messages
+      mails <- liftIO $ mapM messageToMail msgs
+      return $ Vec.fromList mails
+  ) >>= pure . either (Left . show) pure
```
frasertweedale: instead of
```haskell
runExceptT (k) >>= pure . either (Left . show) pure
```

Why not:
```haskell
first show <$> runExceptT (k)
```

(`Data.Bifunctor.first`)

frasertweedale: Also, does `getMessages` run multiple times?  Either we should open the database somewhere else and pass it around, or you should use `databaseDestroy` to close the database after each use (otherwise we leak memory
and probably other resources).

src/Storage/Notmuch.hs
```diff
@@ -1,39 +1,86 @@
...
+    :: FilePath
+    -> NotmuchMail
+    -> IO (Either String NotmuchMail)
+setNotmuchMailTags dbpath m =
+  case mailTagsToNotmuchTags m of
+    Nothing -> pure $ Left "Tags are corrupt"
+    Just nmtags ->
+      bracket (runExceptT (databaseOpen dbpath) >>= either (error . show) pure)
+              (void . runExceptT . databaseDestroy)
+              (\db -> tagsToMessage nmtags m db)
```
frasertweedale: Yeah, `bracket` is a bit awkward because it forces `IO`; the transformer is unhelpful here :/
Not much we can do about it right now.

src/Storage/Notmuch.hs
```diff
@@ -1,39 +1,86 @@
...
+    Left e -> pure $ Left (show e)
+    Right Nothing -> pure $ Left "boop"
+    Right (Just msg) -> do
+      _ <- runExceptT (messageSetTags xs msg)
+      pure $ Right m
+
+addTag :: NotmuchMail -> T.Text -> NotmuchMail
+addTag m t =
+    let xs = view mailTags m
+    in set mailTags (xs `union` [t]) m
```
frasertweedale: Might be nicer to use `over` here (also in `removeTag`)

src/Storage/Notmuch.hs
```diff
@@ -1,39 +1,86 @@
...
+mailTagsToNotmuchTags :: NotmuchMail -> Maybe [Tag]
+mailTagsToNotmuchTags m =
+    let xs = view mailTags m
+    in traverse id (mkTag . encodeUtf8 <$> xs)
 
 messageToMail
-    :: HasTags (Message n RO)
-    => Message n RO
+    :: HasTags (Message n a)
+    => Message n a
```
frasertweedale: 👍

src/UI/Index/Keybindings.hs
```diff
@@ -46,8 +50,25 @@ focusSearch s = continue $ s
...
+updateReadState :: (NotmuchMail -> Text -> NotmuchMail) -> AppState -> IO AppState
+updateReadState op s =
+    case L.listSelectedElement (view (asMailIndex . miListOfMails) s) of
+        Just (_,m) ->
+            let newTag = view (asConfig . confNotmuch . nmNewTag) s
+                dbpath = view (asConfig . confNotmuch . nmDatabase) s
+            in do setNotmuchMailTags dbpath (op m newTag)
+                  >>= either
+                      (\err -> pure $ s & asError ?~ err)
+                      (pure . updateMailInList s)
```
frasertweedale: perhaps the following instead of `(>>=)`
```haskell
either (\err -> set asError (Just err) s) (updateMailInList s)
  <$> setNotmuchMailTags dbpath (op m newTag)
```

src/Storage/Notmuch.hs
romanofski: Many thanks for the review. Yes `getMessages` is used a few times, certainly every time when you changes the search. Would it make sense to open the database with the config?

romanofski: One other thing which comes to my mind is using (resourceT)[https://hackage.haskell.org/package/resourcet] on the client side, to always close the database ... Currently I'd be leaning towards the resourcet approach and always closing the database after reading.

src/Storage/Notmuch.hs
romanofski: oh yes.. much, much better!! Cheers for that.

src/UI/Index/Keybindings.hs
romanofski: Nice. Cheers.

src/Storage/Notmuch.hs
frasertweedale: ResourceT, maybe.  Let's not do it in this rebase.  Something to look at after perhaps.

src/Storage/Notmuch.hs
```diff
@@ -1,39 +1,85 @@
...
-        msgs <- messages q
-        mails <- mapM messageToMail msgs
-        return $ Vec.fromList mails
+getMessages :: T.Text
+            -> NotmuchSettings FilePath
+            -> IO (Either String (Vec.Vector NotmuchMail))
+getMessages s settings =
+  first show <$>
+  bracket (runExceptT (databaseOpenReadOnly (view nmDatabase settings))
+           >>= either (error . show) pure)
```
romanofski: This is still not right IMHO... it would currently crash the application if I wouldn't be able the open the database properly. Arguably that's pretty bad in itself, but perhaps the application shouldn't crash. Same for the function which opens the database RW for tagging. Any suggestions or leave it for now and keep a backlog item.

romanofski:
Alright this should be it. I've added a bunch of fix up patches as well and an acceptance test.

src/Storage/Notmuch.hs
frasertweedale: Leave for now.

Issue #58, opened 2017-09-05
Manipulating the notmuch search is cumbersome


romanofski:
There is a little bug in purebred, which is editing the notmuch search. At this point the user always needs to explicitly move the cursor to the end in order to either remove everything or extend the query. That's frustrating and poor UX.

PR #59, opened 2017-09-05
Place cursor at the end of line

romanofski:
This allows for users to extend or delete the current search terms in an
instand the editor receives focus.

Issue: #58


frasertweedale:
Well I haven't tested it, but it does build and */me looks at code* seem legit.  Merge ho!

PR #61, opened 2017-09-08
Build hs-notmuch against newer notmuch

romanofski:
This installs notmuch locally since the notmuch version available via
system packages is very, very outdated. Furthermore it has a regression
in a macro on which hs-notmuch is dependend upon. Result is failing tests.

To circumvent this problem, install our own supported version of notmuch
and build hs-notmuch against this version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ before_install:
   - mkdir -p ~/.local/bin
   - export PATH=$HOME/.local/bin:/opt/ghc/$GHCVER/bin:$PATH
   - travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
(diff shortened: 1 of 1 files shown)


romanofski:
This now has notmuch 0.24.2 and 0.25 in it. I've ran into intermittend test failures with one test, which is related to tmux. The 0.25 build of hs-notmuch fails tho, but I think you mentioned that.

frasertweedale:
@romanofski right, the v0.25 incompatibility is resolved in the same hs-notmuch commit that changes to using MonadError.  So I'm satisfied with failures against v0.25 until we rebase purebred on tip of hs-notmuch.

romanofski:
Aaah OK. Well I suppose I should rebase the other pull request into this one...

frasertweedale:
@romanofski nah we can merge this first and remove the v0.25 build from expected-failures after the rebase.

PR #63, opened 2017-09-11
dispel some uses of 'error'

frasertweedale:


src/Storage/Notmuch.hs
```diff
@@ -49,22 +51,22 @@ setNotmuchMailTags dbpath m =
...
-      _ <- runExceptT (messageSetTags xs msg)
-      pure $ Right m
+  -- This is nasty.  We have to run the transformer so we can change the
+  -- error type.  Instead, we should update hs-notmuch to abstract over
+  -- an "AsNotmuchError" typeclass, then create our own error sum type
+  -- with an instance of it.
+  runExceptT (
+    findMessage db (view mailId m)
+    >>= traverse (\msg -> messageSetTags xs msg $> m)
+  ) >>= either (throwError . show) (maybe (throwError "boop") pure)
```
romanofski: Heh hahhaha... the "boop" leaked into the outside world. Should have actually been something useful like: Notmuch was not able to find Message with id: 932932329 . Is your notmuch setup working?

src/Storage/Notmuch.hs
```diff
@@ -49,22 +51,22 @@ setNotmuchMailTags dbpath m =
   case mailTagsToNotmuchTags m of
     Nothing -> pure $ Left "Tags are corrupt"
     Just nmtags ->
-      bracket (runExceptT (databaseOpen dbpath) >>= either (error . show) pure)
-              (void . runExceptT . databaseDestroy)
-              (\db -> tagsToMessage nmtags m db)
+      bracket (runExceptT (databaseOpen dbpath))
+              (runExceptT . traverse_ databaseDestroy)
+              (runExceptT . (either (throwError . show) pure >=> tagsToMessage nmtags m))
```
romanofski: Inside the monad transformer you propagate the error by "rethrowing" it?

src/Storage/Notmuch.hs
frasertweedale: Right.  I'll tackle that refactor later, too.

src/Storage/Notmuch.hs
frasertweedale: Essentially yes, because we want to work in MonadError, but `bracket` forces us to `IO`.  So for each "stage" of `bracket` we `runExceptT`, and then in the "main computation" stage we
take the `Either e a` from the resource initialisation stage and "lift" it back into a `MonadError e IO a` to continue the computation.

After I tackle the "new error sum type" refactor discussed in the comment later in commit,
I'll factor out a `MonadError`-friendly variant of `bracket` that lets us avoid doing all of this juggling between `IO (Either e a)` and `MonadError e IO a`.

romanofski: I understand. So merge this and the sum type refactor comes with different patches and separate commits in hs-notmuch, yes?

frasertweedale: Yep.  I'll take that as your ACK :)

PR #64, opened 2017-09-11
Tests dont leak test directories

romanofski:


romanofski:
Little stuff really ...

Issue #60, opened 2017-09-06
NotmuchMail: defer reading the filename until viewing


romanofski:
From our first speed analysis in #51 one aspect which can be improved is to defer the looking up of the filename until it's actually needed. That is when viewing the e-mail. Saving the `message id` is enough to find the e-mail. Doing this for a couple of thousands of records is not necessary.

romanofski:
I've got the patch almost ready for this...

romanofski:
This is merged.

PR #66, opened 2017-09-14
Defer parsing the filename when viewing mail

romanofski:
This gives us a slight speed increase when populating the index for
results with lots of mails. Fixes issue #60.

Issue: #51

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 
@@ -6,6 +7,7 @@ module Storage.Notmuch where
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Except (MonadError, throwError)
+import qualified Data.ByteString as B
 import Data.Traversable (traverse)
 import Data.List (union)
 import Data.Maybe (fromMaybe)
@@ -39,6 +41,14 @@ getMessages s settings =
               mails <- liftIO $ mapM messageToMail msgs
               pure $ Vec.fromList mails
 
+mailFilepath
+  :: (MonadError Error m, MonadIO m)
+  => NotmuchMail -> FilePath -> m FilePath
+mailFilepath m dbpath =
+  bracketT (databaseOpenReadOnly dbpath) databaseDestroy go
+  where
+    go db = getMessage db (view mailId m) >>= messageFilename
+
 setNotmuchMailTags
   :: (MonadError Error m, MonadIO m)
   => FilePath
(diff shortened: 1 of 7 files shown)


romanofski:
Had to chuck away one test for this. Which means this stuff is untested at this point :/

src/Storage/ParsedMail.hs
```diff
@@ -7,18 +7,21 @@ import Codec.MIME.Parse (parseMIMEMessage)
...
-        Left e -> pure $ Left $ show e
-        Right contents -> pure $ Right $ MIMEMail (parseMIMEMessage contents)
+parseMail :: NotmuchMail
+          -> FilePath
+          -> IO (Either String ParsedMail)
+parseMail m dbpath =
+    mailFilepath m dbpath
+           >>= either fail (\x -> try (readFile x):: IO (Either IOError T.Text))
+           >>= either (\x -> pure $ Left (show x <> " Maybe run notmuch new again?")) (pure . Right)
+           >>= either (pure . Left . show) (pure . Right . MIMEMail <$> parseMIMEMessage)
```
frasertweedale: `fail` is bad bad bad.

Do this instead:
```haskell
parseMail m dbpath =                                                                                     
  mailFilepath m dbpath                                                                                  
  >>= fmap (bimap (\x -> show x <> " Maybe run notmuch new again?") (MIMEMail . parseMIMEMessage) . join)
      . traverse (fmap (first (show :: IOError -> String)) . try . T.readFile)                           
```

The whole thing will improve once I do the "abstract over error type" refactor, though.

src/Storage/Notmuch.hs
```diff
@@ -34,15 +34,34 @@ getMessages :: T.Text
...
+mailFilepath m dbpath =
+    withRODatabase dbpath go >>=
+    either
+        (pure . Left . show)
+        (pure .
+         maybe
+             (Left "No message found. Is your notmuch correctly setup?")
+             Right)
+  where
+    go db = findMessage db (view mailId m) >>= traverse messageFilename
```
frasertweedale: can tidy this one up a bit:
```haskell
mailFilepath :: NotmuchMail -> FilePath -> IO (Either String FilePath)              
mailFilepath m dbpath =                                                             
    join . bimap show (maybe (Left notFoundMsg) Right) <$> withRODatabase dbpath go 
  where                                                                             
    notFoundMsg = "No message found. Is your notmuch correctly setup?"              
    go db = findMessage db (view mailId m) >>= traverse messageFilename             
```

Again, the forthcoming refactoring should improve things a bit more.

frasertweedale:
@romanofski can we get a proper rebase and conflict resolution for this PR?

romanofski:
hm... what the hell...

frasertweedale:
@romanofski I think this is good to do but can you please build and run it as a sanity check? :)

Seriously I never run the program, I just refactor stuff and send PRs :)

romanofski:
Yep working fine. There is already a test which covers viewing mail, so in case something would have gone south you would have noticed it. With the test re-factor I'll have the possibility to add in a test which also checks the error handling. At least the case in which the mail file is gone, but it's still indexed.

PR #70, opened 2017-09-27
Error sum type refactor

frasertweedale:
Once this is landed I'll rebase #66.


frasertweedale:
On Wed, Sep 27, 2017 at 11:46:08PM -0700, Roman Joost wrote:
> romanofski approved this pull request.
> 
> Aaah I see what you did here. Nice. Should have thought about this myself, since it's not very complicated.
> 
well, it was worth waiting for the hs-notmuch error refactor before
attempt thing anyway.

PR #71, opened 2017-09-28
Declarative acceptance tests

romanofski:
This is currently still work in progress, but I've got a few questions, so take it as an invite for feedback.

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -94,3 +94,5 @@ test-suite unittests
                      , resourcet
                      , time
                      , regex-posix
+                     , mtl
+                     , lens
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -20,7 +20,7 @@ import Control.Lens (view, _3, _2)
 import Data.List (isInfixOf)
 import System.Process (callProcess, readProcess)
 import System.Directory
-       (getCurrentDirectory, removeDirectoryRecursive)
+       (getCurrentDirectory, removeDirectoryRecursive, removeFile)
 import Test.Tasty (TestTree, TestName, testGroup, withResource)
 import Test.Tasty.HUnit (testCaseSteps, assertBool)
 import Text.Regex.Posix ((=~))
@@ -34,7 +34,21 @@ systemTests =
         , testUserCanManipulateNMQuery
         , testUserCanSwitchBackToIndex
         , testCanToggleHeaders
-        , testSetsMailToRead]
+        , testSetsMailToRead
+        , testErrorHandling]
+
+testErrorHandling ::
+  TestTree
+testErrorHandling = withTmuxSession "error handling" $
+  \step -> do
+    startApplication
+
+    testmdir <- getTestMaildir
+    liftIO $ removeFile (testmdir <> "/new/1502941827.R15455991756849358775.url")
+
+    liftIO $ step "shows error message"
+    out <- sendKeys "Enter" "FileReadError"
+    assertSubstrInOutput "openFile: does not exist" out
 
 testSetsMailToRead ::
   TestTree
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 {-# LANGUAGE OverloadedStrings #-}
 module TestUserAcceptance where
@@ -6,26 +7,27 @@ import qualified Data.Text as T
 import System.IO.Temp (createTempDirectory, getCanonicalTemporaryDirectory)
 import Data.Ini (parseIni, writeIniFileWith, KeySeparator(..), WriteIniSettings(..))
 import Data.Semigroup ((<>))
-import Control.Concurrent
-       (newEmptyMVar, putMVar, takeMVar, MVar, threadDelay)
+import Control.Concurrent (threadDelay)
 import Control.Exception (catch, IOException)
 import System.IO (hPutStr, stderr)
 import System.Environment (lookupEnv)
 import Control.Monad (void, when)
 import Data.Maybe (isJust)
+import Control.Monad.IO.Class (liftIO)
+import Control.Monad.Reader (runReaderT, ask, ReaderT)
 
+import Control.Lens (view, _3, _2)
 import Data.List (isInfixOf)
 import System.Process (callProcess, readProcess)
 import System.Directory
        (getCurrentDirectory, removeDirectoryRecursive)
-import Test.Tasty (TestTree, testGroup, withResource, mkTimeout, localOption)
-import Test.Tasty.HUnit (testCaseSteps, assertBool, Assertion)
+import Test.Tasty (TestTree, TestName, testGroup, withResource)
+import Test.Tasty.HUnit (testCaseSteps, assertBool)
 import Text.Regex.Posix ((=~))
 
 systemTests ::
   TestTree
 systemTests =
-    localOption (mkTimeout testTimeout) $
     testGroup
         "user acceptance tests"
         [ testUserViewsMailSuccessfully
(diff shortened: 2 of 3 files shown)


purebred.cabal
```diff
@@ -91,3 +91,5 @@ test-suite unittests
                      , resourcet
                      , time
                      , regex-posix
+                     , mtl
+                     , lens
```
romanofski: Not sure if that is a bit over the top really just to pluck out the third element of a tuple. Maybe should use a data type?

src/Storage/Notmuch.hs
```diff
@@ -1,16 +1,15 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE LambdaCase #-}
 -- | module for integrating notmuch within purebred
 module Storage.Notmuch where
```
romanofski: Sorry ignore this ...

test/TestUserAcceptance.hs
```diff
@@ -295,33 +242,46 @@ cleanUpTmuxSession sessionname = do
...
 
 holdOffTime :: Int
 holdOffTime = 10^6
 
+debugOutput :: String -> IO ()
+debugOutput out = do
+  d <- lookupEnv "DEBUG"
+  when (isJust d) $ hPutStr stderr ("\n\n" ++ out)
+
+wait :: String -> String -> IO (String)
```
romanofski: Currently the tests make a screen shot for each step. I'll loose that with this refactor. Although, it's rather just a matter of putting it somewhere else.

One thing which can be frustrating to debug is when an action times out, because whatever string we were waiting for to appear didn't appear (crashed, different characters, whatever). Having a dump in that instance is really handy. So, loosing the fact that we don't have screenshots for each passing step is not much of a big deal, if I can make one in an instance when the wait function times out.

The `waitForString` function currently invokes itself recursively until it times out. I wonder if I can refactor this and only wait like 5 * the hold off time, raise an assertion error and dump the screen. If something jumps into your mind I'd be all ears.

test/TestUserAcceptance.hs
```diff
@@ -42,198 +49,138 @@ testTimeout = 10 ^ 6 * 8
...
-assertSubstrInOutput out substr = assertBool "in out" $ substr `isInfixOf` out
+  withTmuxSession "user can switch back to mail index during composition" $
+        \step -> do
+            liftIO $ step "start composition"
+            out <- sendKeys "m" "From"
+            assertSubstrInOutput "From" out
+
+            liftIO $ step "enter from email"
+            out <- sendKeys "testuser@foo.test\r" "To"
+            assertSubstrInOutput "To" out
```
romanofski: Once refactored, I've actually realised that all these assertions are kind of superfluous since the `sendKeys` function already waits for the string I'm asserting against to appear. Perhaps an improvement another day.

frasertweedale:
There are conflicts.  And can you squash all the WIP commits?

romanofski:
Will do.

romanofski:
I've got an idea how to refactor the `waitForString` function, but it isn't nice. Basically use an additional parameter (a counter) and bail out once it reached the max amount of recursive calls I've reached. Additionally instead of just using a dumb baton as `MVar`, I use an `Int` and then assert that it's positive, otherwise the failure dumps the screen.

It "works" but.. hm... feels convoluted. Maybe I should let is sit for a while and think about it.

test/TestUserAcceptance.hs
```diff
@@ -295,59 +242,69 @@ cleanUpTmuxSession sessionname = do
...
+  assertBool ("expected: '" ++ substr ++ "' but got:\n\n " ++ out) (i == 1)
+  pure out
+
 -- | wait for the application to render a new interface which we determine with
 --   a given substring. If the expected substring is not in the captured pane,
 --   wait a bit and try again.
-waitForString :: MVar String -> String -> String -> IO (String)
-waitForString baton sessionname substr = do
+waitForString :: MVar Int -> String -> String -> Int -> IO (String)
+waitForString baton sessionname substr i = do
```
frasertweedale: I think it would be better to make the Int arg a "countdown", and put the number of desired attempts at the call site.
So something like:
```haskell
waitForString :: Int {- attempts -} -> MVar Int -> String -> String -> IO (String)
waitForString n baton sessionname substr =
  readProcess "tmux" ["capture-pane", "-e", "-p", "-t", sessionname] [] >>= checkPane
  where
    checkPane out
      | substr `isInfixOf` out = putMVar baton 1 $> out
      | n <= 0 = putMVar baton 0 $> out                     -- no more attempts
      | otherwise = threadDelay holdOffTime *> waitForString baton sessionname substr (n - 1)
```

Personally I think this is easier to follow with the guards.

Then you would call it like:
```haskell
  out <- waitForString baton sessionname substr 5 {- max tries -}
```

If you ever do need to get back the "screenshot for each attempt" it is a straightforward refactor;
you just accumulate the screenshots in a list or something.

purebred.cabal
frasertweedale: nah go for it mate.  we're already depending on these in the lib/main program so it's no biggie.

test/TestUserAcceptance.hs
frasertweedale: Yeah whatevs mate, I got no idea; tests are your department :)

test/TestUserAcceptance.hs
```diff
@@ -295,59 +242,69 @@ cleanUpTmuxSession sessionname = do
...
     out <- readProcess "tmux" ["capture-pane", "-e", "-p", "-t", sessionname] []
```
frasertweedale: Can we refactor this to use `capture :: ReaderT Env IO (String)`?  The `readProcess` logic is duplicated.

test/TestUserAcceptance.hs
```diff
@@ -42,198 +49,138 @@ testTimeout = 10 ^ 6 * 8
...
+            assertSubstrInOutput "Testmail" out
+
+            liftIO $ step "switch back to the compose editor"
+            out <- sendKeys  "Tab" "test subject"
+            assertSubstrInOutput "test subject" out
+
+type Env = (String, String, String)
+
+assertSubstrInOutput :: String -> String -> ReaderT Env IO ()
+assertSubstrInOutput substr out = liftIO $ assertBool (substr ++ " not found in\n\n" ++ out) $ substr `isInfixOf` out
```
frasertweedale: If we're touching these lines anyway, can we replace applications of `(++)` with
`Data.Semigroup.<>`?

Just a nit, I don't really mind if we don't do it :)

test/TestUserAcceptance.hs
```diff
@@ -13,13 +14,19 @@ import System.IO (hPutStr, stderr)
 import System.Environment (lookupEnv)
 import Control.Monad (void, when)
 import Data.Maybe (isJust)
+import Control.Monad.IO.Class (liftIO)
+import Control.Monad.Reader (runReaderT, ask, ReaderT)
 
+import Control.Lens.Getter (view)
+import Control.Lens.Tuple (_3)
```
frasertweedale: Can we just import `view` and `_3` from the top-level module?

```haskell
import Control.Lens (_3, view)
```

romanofski: Ah... did not know ... cheers!

test/TestUserAcceptance.hs
romanofski: Good point. Will do.

test/TestUserAcceptance.hs
romanofski: AAahh guards... of course. Why haven't I thought of them. Cheers! I like the countdown.

test/TestUserAcceptance.hs
romanofski: Hm.. tricky. I thought I could pull it of with also changing `waitForString` using `ReaderT`, but the `otherwise` branch then ends up being lifted in IO and I miss out recursively calling `waitForString` again... Will have another look at it tomorrow.

romanofski:
Now I think I've got it working again... I'm currently fiddling with the timing. I might increase the amount to poll for the terminal to re-draw and maybe remove the `tasty` timeout option overall.

romanofski:
Alright finally I think I've got it all together. When adding the test for the error handling I've realized that we currently have no good way to clear the error. Currently errors are only cleared if Keybindings don't match and the default handler handles the key press. I'll have to put that in a separate backlog item :/

frasertweedale:
Aight let's merge this.

PR #72, opened 2017-09-28
add explicit type signatures to HLint annotations

frasertweedale:
When loading the modules in GHCi with -XOverloadedStrings, the type
of the annotation is ambiguous.  Add explicit type annotations so
the modules can be loaded in the REPL.

Also add a note to HACKING about including this annotation.


romanofski:
Good stuff. I'm running hlint to little (thought it was automatically run with flycheck, but doesn't seem to be the case).

Issue #73, opened 2017-09-29
Clearing errors


romanofski:
While errors are displayed in the UI, getting them cleared is currently not working as expected. They stick around until the user would press a key which is not handled by the key bindings.

Initially I had in mind a behaviour like this for example:

* user wants to read an e-mail which doesn't exist
* purebred shows an error
* user moves on to the next mail or anywhere else in the index and the error is removed

Clearing the error for the next keyboard input purebred receives could mean, that the user accidently "scrolls" over the error and misses it. Yet at this point in time, I'd stick to the basics and propose to clear any error on each next keyboard input we receive.

PR #74, opened 2017-10-01
Refactor keybindings

romanofski:


src/UI/Keybindings.hs
```diff
@@ -1,35 +1,66 @@
...
+handleKeyboardEvent s (T.VtyEvent ev) =
+    let mode = view asAppMode s
+        kbs = fromMaybe [] $ view (asConfig . confKeybindings . at mode) s
+    in case lookupKeybinding ev kbs of
+           Just kb -> view kbAction kb s
+           Nothing -> fallbackKeybindingHandler mode s ev
+handleKeyboardEvent s _ = M.continue s
+
+
+fallbackKeybindingHandler :: Mode
```
romanofski: So what this simplifies is the overall registration of the key bindings into a central place. What I can not get around is how I should model some kind of fall through mode (like in pattern matching how something matches all). I do this here with this `fallbackKeybindingHandler` but it simply attributes for the fact, that some keys should be allowed to be forwarded to the widget.

Simply put, all keybindings work such that we forward all keys to the widget which has the focus by the mode (currently hard coded), except what is defined in the keybinding.

For example the editor is a prime example: all keys should be forwarded to the editor (since we want to type something), except the ones defined in the key bindings. Is this way perhaps acceptable or do we need something better? At least it seems the most expected when re-configuring the keybindings in the configuration.

Also, looking at #73 I suppose just handling the action by a keybinding is not enough. We need something which can come after the action. Arguably this could be looked at when working on #73 .

romanofski:
I think this could wait until the declarative tests pull request is merged. I could think of adding a test case which tests that a key binding can be redefined with our configuration or something else ...

romanofski:
Based on this, I've got a fix for #75 and I think I've been over thinking it a bit with my previous comment. Trying out how to configure new key bindings, I think this is at least better than what we have at this point.

romanofski:
Alright, if you've got time would you have a peek at e6ad31b2d72ec8731fc02cd1e9f3c2f99ab0d411 and check if that is what you had in mind? So basically it's more or less shifting around with the exception that the Keybinding now also gets the widget type. But this type is not passed as a parameter to the action or anything, but rather used for ... type checking?

frasertweedale:
@romanofski yeah overall that commit is good mate.  I'm happy for it to be merged after you add
the export list.

romanofski:
@frasertweedale yeah it's just a beginning, but I thought better to figure out that it's the wrong direction than moving everything around and it's for the bin. Cheers.

romanofski:
Actually now realising this - and maybe thats ok - we do end up with separate lists for each widget. For example the config has two keybinding lists for the index view: `_ivKeybindings`, `_ivSearchKeybindings`. For the mailview we will have to have the same, since I can't mix Keybindings abstracted over a list (the index is still shown, even tho it's smaller) and the widget showing the mail.
Not sure how this may pan out.. just saying ...

romanofski:
Hm... hitting a road block here. So far re-factoring the index keybindings works splendidly as well as re-using them in the mail view (esp. the ones to go back to browsing e-mail). The tricky bit is implementing the more specific scrolling actions, since I think I can't find a common denominator. First an foremost I'm creating a new instance for each rendering step of the scrolling widget without keeping the widget itself in the application state. I think that's fine, since for the scrolling itself I just need to know the widget name.

The action:

```Haskell
data Action a = Action
    { _aDescription :: String
    , _aAction :: AppState -> EventM Name (Next AppState)
    }
```
does not keep any information about a widget or its name. That is good, because otherwise I wouldn't be able to define general actions such as going back to the index.

What works for general actions however, doesn't work for the scrolling tho, since in the action there is no reference to the view name.

I could implement the specific scrolling instance for a `Widget Name` and inspect the application state like:

```Haskell
instance Scrollable (T.Widget a) where
  viewport s = case view asAppMode s of
      ViewMail -> viewportScroll ScrollingMailView
      _ -> -- ???? what there?
```
or I use the name:

```Haskell
instance Scrollable (T.Widget a) where
  viewport _ = viewportScroll ScrollingMailView
```
but that makes no sense either.

Maybe you can see what I can't see at the moment. Perhaps I have to back up and think about it again ... Pushed my current state to 574eec5dc9f6440f3f87398ff8a2beb9ed82848d for reference.

romanofski:
Hmpf.. now I thought instead of the widget, I abstract over the `Name`, since that's the identifier for the widget and for some widget, I can grab the [name](https://hackage.haskell.org/package/brick-0.20.1/docs/Brick-Widgets-Core.html#v:getName). But that won't work either, since at some point I'll have to provide the name constructor to the action. Furthermore the whole point of using the widget was to avoid registering pointless actions and with the name as a type we're back at square one.

I think in essence it comes down to how do I pass the `Name` constructor to the action. I first thought I could do this with the class constraint, but that seems not to be working out as I thought, since all I would have is one single instance which would guess the name based on the `Mode` in the application state :/

romanofski:
I've now converted all Keybindings and realised another inconsistency within Purebred I personally put in. I thought I would always know based on the application state which widget has the focus. That is not entirely true, since for the view showing the mail while allowing to select the next mail it currently is just a hardcoded fallback which diverts all input to the list of mails. Ideally I should be able to almost re-use all actions I had defined for the index of mails.

romanofski:
@frasertweedale maybe it's something you could help with.

So for viewing mail, I'd like to compose two actions e.g.:

```Haskell
Keybinding (V.EvKey V.KDown []) (mailIndexDown >>= displayMail)
```
Tho then I get the error that there is no instance for `Monad Action`. The actual actions type is `AppState -> EventM Name (Next AppState)` and since it's a function I'm not sure if:

a) implementing `Monad Action` is the right thing here to do
b) should I be able to compose each of those two action functions?

romanofski:
pushed my latest head in the split_keybindings branch (dfa0dc40565bcdc07023a246040393d6ae9f9ed5)

frasertweedale:
@romanofski yeah I'll have to think about how to compose actions.  I see what you're
trying to achieve there I just don't know how the types line up yet.  It will be possible but not
as-is with `Monad`.

frasertweedale:
@romanofski if everything is working with the current tip of this PR, let's merge it.

frasertweedale:
@romanofski So for composing actions together into a "this then that" action, there are some things to consider:

1. state from first action must be passed to second action

2. the names of the actions must be joined some how (basically a string concat)

3. the `Next a` type is *opaque* so there is actually nothing we can do to chain
    an `EventM Name (Next AppState)` to a subsequent action.

So what do we do?

Why, we abstract some more!  And write one or two combinators.

Rough idea:

```haskell
data Action ctx a = Action {- abstract over widget context AND action return type -}
    { _aDescription :: String
    , _aAction :: AppState -> EventM Name a
    }

continue :: Action ctx (Next AppState)
continue = Action "" Brick.continue

-- Run an action that returns a **pure** AppState,
-- then feed the result into a second action.
--
-- Using this, actions that result in changed `AppState`
-- can be easily chained together, e.g.:
--
--    Keybinding (V.EvKey V.KDown []) (mailIndexDown `chain` displayMail `chain` continue)
--
chain :: Action w AppState -> Action w b -> Action w b
chain (Action s1 f1) (Action s2 f2) =
  Action
    (if null s2 then s1 else s1 <> " and then " <> s2)
    (f1 >=> f2)
```

This will require refactoring a lot of actions to return a pure `AppState` instead of
`Next AppState`.  And updating various keybinding definitions to finish a chain of
actions with (our) `continue`.

romanofski:
I know this turns out to be a bit like a rabbit hole, but what the heck. Let me try this as a separate patch. It seems interesting ...

frasertweedale:
Uh... I dont' think I was supposed to merge this... we didn't want to go for the 'global map' approach did we?

frasertweedale:
I reverted the tip.

romanofski:
Heh yeah sorry. I wasn't sure myself when we took a turn what to do with this PR ...

romanofski:
There is no way to re-open this one is there. Ah well, I'll create a new one with the refactoring ...

PR #76, opened 2017-10-12
Bump to latest brick release

romanofski:
Due to failing test failures with brick 0.20 and the latest incompatible
vty release which is picked, we bump our brick dependency to the latest
and greatest.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -46,9 +46,10 @@ packages:
 # (e.g., acme-missiles-0.3)
 extra-deps:
 - mime-0.4.0.2
-- brick-0.20.1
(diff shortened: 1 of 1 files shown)


romanofski:
Not Keybindings, but my branch ran into trouble against stack nightly so I needed to do something about that first.

frasertweedale:
Is there a way to make the stackage nightly jobs "allowed to fail"?

frasertweedale:
The commit is green, the merged PR is not.  Ima merge it anyway since there's clearly nothing
wrong with this PR on its own.

PR #77, opened 2017-10-12
Some minor refactoring of event handling

frasertweedale:


romanofski:
You'll most likely need to rebase against #79 for the tests to pass :)

romanofski:
In regards to the patch:

I suppose we're currently not supporting mouse events anyway so the change is fine with me.

frasertweedale:
Other kinds of events we might handle in the future could be custom events coming from e.g. a
thread monitoring for new mail or something like that.  But we can deal with the implications of that if/when we are actually implementing it :)

Issue #78, opened 2017-10-12
tests: terminal control sequences extracted from tmux can differ from expectations


frasertweedale:
Not quite sure why this situation arises, but my tests are failling like so:

```
  user acceptance tests                                                                                  
    user can view mail:                                       OK (2.04s)                                 
      shows tag                                     
      view mail                                     
    manipulating notmuch search query results in empty index: FAIL (6.05s)                               
      focus command                                 
        Wait time exceeded. Expected: '37;40mtag' last screen shot:                                      
                                                    
          <roman@url.user 17/Aug  Testmail                                          inbox                
         <rjoost@url.use 16/Aug  This is Purebred                          inbox replied                 
                                                    

        Purebred: Item 0 of 2                       
        tag:inbox                                   
                                                    
                                                    
         raw: "\ESC[1m\ESC[37m\ESC[43m <roman@url.user 17/Aug  Testmail                                          \ESC[0m\ESC[36m\ESC[43minbox\n\ESC[94m\ESC[40m <rjoost@url.use 16/Aug  This is Purebred           
               \ESC[36minbox replied\n\ESC[94m\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\ESC[30m\ESC[mPurebred: Item 0 of 2\n\ESC[37m\ESC[40mtag:inbox\n"
```

Above you will see that I added a raw print of the string captured from tmux, to
inspect the raw escape codes.

As you can see, before the NM search widget there is

    \ESC[37m\ESC[40m

instead of what the search is looking for:

    \ESC[37;40m

Semantically these are the same, but as I said, for whatever reason (maybe some weird
terminfo or environment settings) the codes being sent are different.

As a first step to resolving this, I'd propose exfiltrating `$TERM` or other terminal information
from the tmux session and reporting that somehow, to see what might differ.  In my environment
`TERM=screen-256color`

UPDATE: this actually seems to be related to tmux only.  When I write ASNI colour codes to term
using semicolon then *capture-pane* tmux reports them in the way described about.  It must be
something to do with my tmux version or settings.

PR #80, opened 2017-10-13
fix tests with tmux >= 2.5

frasertweedale:
In tmux < 03d01eabb5c5227f56b6b44d04964c1328802628 (first released
in tmux-2.5), the `capture-pane -e' output ran attributes,
foreground colour and background colour params together into one
escape sequence.

After that commit, attributes, foreground colours and background
colours are written in separate escape sequences.  Therefore for
compatibility with different versions of tmux we need to check the
ANSI escape sequences in two different ways.

Fixes: https://github.com/purebred-mua/purebred/issues/78


frasertweedale:
Nevermind that I broke master... if all the other tests still work after this commit, let's merge
it :)

romanofski:
I like the change :)

Issue #45, opened 2017-08-26
How to address the configuration of keybinding


romanofski:
At this point Keybindings are configured as a flat list in each view in separate configuration slots. Configuring different Keybindings turns out to be tricky, since the user:

* needs to import all Keybindings for the specific view he wants to change the bindings for
* needs to know the list operations to merge his Keybindings with the ones which come as default

The problem with that approach is, if you stuff the merging of your and the default key bindings, you end up with the default keybindings which are not necessarily the ones you know (yes there should be a view which shows all of them, but that's a separate issue).

Since we're already recording which view has the focus, I'm thinking instead of configuring the key bindings as a list for each view, why not configure the bindings as one tree.

* each branch is basically the focus of the current view
* it will allow to traverse up (e.g. composing a mail, back to the index view)
* it will allow to easier "merge" the users with the default bindings
* it should significantly make it easier to look up key bindings

So I can think of a lot of benefits with the new approach, but no real draw backs come to my mind at the moment. Perhaps it's not as peachy as I envision it...

PR #82, opened 2017-10-13
Refactor Keybindings

romanofski:
Split up the Action in to it's own datatype from the binding itself.
The motivation is to compose multiple actions if there is need.

For example, to move the cursor up and down the index we can define one
action: moveIndexUp. Yet we can re-use the same action with an action to
update the current view with a parsed mail when we actually view
e-mails.

Issue: #45

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -3,7 +3,8 @@ module Types where
 
 import Codec.MIME.Type (MIMEValue)
 import qualified Brick.AttrMap             as Brick
-import           Brick.Types               (EventM, Next)
+import Brick.Types (EventM, Next, Widget)
+import Brick.Main (viewportScroll, ViewportScroll)
 import qualified Brick.Widgets.Edit        as E
 import qualified Brick.Widgets.List        as L
 import           Control.Lens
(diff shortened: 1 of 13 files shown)


romanofski:
This doesn't include the discussed re-factoring of the context to allow composing multiple actions. I would leave that for a separate patch set.

src/Types.hs
```diff
@@ -262,3 +273,10 @@ mailTags = lens _mailTags (\m t -> m { _mailTags = t })
 
 mailId :: Lens' NotmuchMail ByteString
 mailId = lens _mailId (\m i -> m { _mailId = i })
+
+class Scrollable n where
+  makeViewportScroller :: AppState -> ViewportScroll n
+
+-- XXX this is currently our only scrolling view
```
frasertweedale: the upcoming "list of key bindings" i.e. "help" view will be scrollable too, I guess

src/UI/Actions.hs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+haltApp :: Action a
+haltApp =
+    Action
+    { _aDescription = "quit the application"
+    , _aAction = halt
+    }
+
+composeMail :: Action (L.List Name NotmuchMail)
```
frasertweedale: This implies that we can only begin composing a mail from the mail index?
Fine for now - I'm just checking my mental model and suggesting that we might
want to generalise it **at a future point**.

src/UI/Actions.hs
```diff
@@ -0,0 +1,246 @@
...
+switchComposeEditor :: Action (L.List Name NotmuchMail)
+switchComposeEditor =
+    Action
+    { _aDescription = "switch to compose editor"
+    , _aAction = \s -> case view (asCompose . cTmpFile) s of
+                          Just _ -> continue $ set asAppMode ComposeEditor s
+                          Nothing -> continue s
+    }
+
+replyMail :: Action (L.List Name NotmuchMail)
```
frasertweedale: Does this action apply in the "view mail" mode too? Should it?  Again, not a blocker for
this PR.

src/Types.hs
romanofski: Yes it will. Tho, isn't this single instance for `Scrollable n` pointless?

src/UI/Actions.hs
romanofski: Yes currently that's the case. I ran into a situation today tho, that I think generalising it would make sense.

src/UI/Actions.hs
romanofski: Yeah it doesn't, but good point.

PR #83, opened 2017-10-13
Abstract Action over pure AppState

romanofski:
In order to chain actions together to a) reduce duplication and b) allow
to compose multiple actions this re-factoring adds an additional
abstraction over the AppState.

Issue: #45

### DIFF
--- a/src/UI/Mail/Main.hs
+++ b/src/UI/Mail/Main.hs
@@ -8,23 +8,17 @@ import qualified Brick.Types as T
 import Brick.Widgets.Core
        (padLeft, padTop, txt, txtWrap, vLimit, viewport, (<+>), (<=>),
         withAttr)
-import qualified Brick.Widgets.List as L
 
 import Codec.MIME.Type
        (MIMEContent(..), MIMEParam(..), MIMEValue(..), Type(..),
         showMIMEType)
 import qualified Data.CaseInsensitive as CI
 import Control.Lens.Getter (view)
-import Control.Lens.Setter (set)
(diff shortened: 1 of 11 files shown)


romanofski:
This worked out wonderfully. I did not incorporate some of the points you raised in the previous PR, since composing mail while viewing mail should also be accompanied with a test. Did not feel that I would also want to cram a few tests into this PR so...

frasertweedale:
@romanofski yeah yeah, all good.  Keep separate things separate, and changes small as possible while still self-contained.  Will review this one some time over weekend.

src/UI/Mail/Main.hs
```diff
@@ -8,23 +8,18 @@ import qualified Brick.Types as T
...
 
 import Codec.MIME.Type
        (MIMEContent(..), MIMEParam(..), MIMEValue(..), Type(..),
         showMIMEType)
 import qualified Data.CaseInsensitive as CI
 import Control.Lens.Getter (view)
-import Control.Lens.Setter (set)
 import Data.CaseInsensitive (mk)
-import Control.Monad.IO.Class (liftIO)
+import Data.Maybe (maybe)
```
frasertweedale: isn't `maybe` in `Prelude`?

frasertweedale:
If we can do without that `maybe` import please fix it up.  Apart from that I'm happy for you push this as-is.

src/UI/Mail/Main.hs
romanofski: Heh yeah good point. Will remove it.

PR #84, opened 2017-10-14
Show help

romanofski:


romanofski:
Rebased against the keybindings pull request. Not sure how GitHub handles it, but we'll see.

romanofski:
Hm there is one test which seems to be failing randomly. It's not that random tho, since it's only one single test. Will have to investigate.

frasertweedale:
@romanofski yeah, I have seen that failure once, too.  Maybe change to invoking the process in a way that we can see stdout and stderr so we can hopefully find more info about it.

romanofski:
Just re-based this patch. Will have another look at changing the invocation for debugging purposes.

frasertweedale:
It works, and I see no reason not to merge it.  But I propose the following enhancements,
which can be addressed separately later:

- the keybindings for "current view" i.e. the view that was active just prior to showing the help, should be shown at the top

- that seems to imply a notion of a "stack" of views, and a "close" or "go back" action

- this seems to suggest a notion of "common" or "universal" keybindings i.e. sets of bindings that are implicitly added to each view.  A similar notion might exist for scrolling, i.e. set scrolling keybindings as a separate config, and those bindings get merged with the view-specific bindings for all "scrollable" views.

frasertweedale:
But as for this patch, it does what it says on the tin, so merge ho!

romanofski:
Put your ideas into #85

PR #87, opened 2017-10-23
UI: Show the mode in the status

romanofski:
This is a little debugging tweak. I can foreseen that we will refactor
more code and abstract it over the mode. Having this is currently very
handy.


romanofski:
Nothing fancy, but I thought it's currently a good way for debugging. I can foresee that we want to keep a persistent status bar and change content while being in different modes :)

PR #88, opened 2017-10-24
Mail add tags

romanofski:


romanofski:
Alright.. this brings us a step away from vaporware :-)

src/UI/Actions.hs
```diff
@@ -237,10 +331,9 @@ updateMailInList m s =
 setError :: Error -> AppState -> AppState
 setError = set asError . Just
 
-reloadListOfMails :: Vector NotmuchMail -> AppState -> AppState
-reloadListOfMails vec =
-  set (asMailIndex . miListOfMails) (L.list ListOfMails vec 1)
-  . set asAppMode BrowseMail
+updateMailsList :: Vector NotmuchMail -> AppState -> AppState
```
romanofski: I think this is buggy. If you carry out a search which yields no results and an additional search on top of it which yields results, the current selected index is Nothing.

src/UI/Actions.hs
```diff
@@ -236,32 +270,58 @@ toggleHeaders = Action
...
 applySearch :: AppState -> T.EventM Name AppState
 applySearch s =
    runExceptT (getMessages searchterms (view (asConfig . confNotmuch) s))
-   >>= pure . ($ s) . either setError reloadListOfMails
+   >>= pure . ($ s) . either setError updateMailsList
      where searchterms = currentLine $ view (asMailIndex . miSearchEditor . E.editContentsL) s
 
+selectedMailHelper
+    :: Applicative f
+    => AppState -> (NotmuchMail -> f (AppState -> AppState)) -> f AppState
```
frasertweedale: Not for now, but I see this as being generalisable to `selectedItemsHelper`.  The next time we need a helper to do something with a "selected item(s)", remember this :)

test/TestUserAcceptance.hs
```diff
@@ -46,8 +46,46 @@ systemTests =
         , testSetsMailToRead
         , testErrorHandling
         , testHelp
+        , testManageTags
         ]
 
+-- TODD: change search -> tag -> search is what was saved
```
frasertweedale: who is Todd?  ahah

configs/config.hs
```diff
@@ -15,11 +16,15 @@ myIndexKeybindings =
     , Keybinding (EvKey KUp []) (mailIndexUp `chain` continue)
     , Keybinding (EvKey (KChar 'k') []) (mailIndexUp `chain` continue)
     , Keybinding (EvKey (KChar '\t') []) (switchComposeEditor `chain` continue)
+    , Keybinding (EvKey (KChar 'a') []) ((applyTags ["archive"]) `chain` reloadMails `chain` continue)
```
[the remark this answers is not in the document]
romanofski: Yup. Bad example? Reminds me to check it's we have an add and remove action. Maybe we should also expose set operations? Hm although except union and difference, not that I can imaging all of that have I had it in a config.

src/UI/Actions.hs
romanofski: Good idea actually.

romanofski: Good idea actually.

test/TestUserAcceptance.hs
romanofski: Doh

src/UI/Actions.hs
```diff
@@ -271,10 +331,9 @@ updateMailInList m s =
 setError :: Error -> AppState -> AppState
 setError = set asError . Just
 
-reloadListOfMails :: Vector NotmuchMail -> AppState -> AppState
-reloadListOfMails vec =
-  set (asMailIndex . miListOfMails) (L.list ListOfMails vec 1)
-  . set asAppMode BrowseMail
+updateMailsList :: Vector NotmuchMail -> AppState -> AppState
+updateMailsList vec s = let current = view (asMailIndex . miListOfMails . L.listSelectedL) s
+                        in over (asMailIndex . miListOfMails) (L.listReplace vec current) s
```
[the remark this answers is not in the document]
romanofski: Hm... not sure if it's actually a bug in [brick](https://hackage.haskell.org/package/brick-0.28/docs/src/Brick-Widgets-List.html#listReplace). It has a fail safe for the selected index if the vector is empty or the given index is out of bounds. If you invoke it with a non-empty Vector, but a `Nothing` index it's selecting nothing. Technically it seems it's doing the right thing?

frasertweedale:
On Tue, Oct 24, 2017 at 04:15:44AM -0700, Roman Joost wrote:
> Hm... not sure if it's actually a bug in [brick](https://hackage.haskell.org/package/brick-0.28/docs/src/Brick-Widgets-List.html#listReplace). It has a fail safe for the selected index if the vector is empty or the given index is out of bounds. If you invoke it with a non-empty Vector, but a `Nothing` index it's selecting nothing. Technically it seems it's doing the right thing?
>
Dunno mate, you are the brick expert :)

src/UI/Index/Main.hs
```diff
@@ -30,7 +30,7 @@ import Config.Main
 drawMain :: AppState -> [Widget Name]
 drawMain s = [ui]
   where
-    editorFocus = view asAppMode s == SearchMail
+    editorFocus = view asAppMode s == SearchMail || view asAppMode s == ManageTags
```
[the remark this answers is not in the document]
romanofski: Done

Issue #90, opened 2017-10-25
README: add screenshot


frasertweedale:
Add a screenshot to the README, so people may bask in the majesty of our terminal UI.

PR #91, opened 2017-10-25
Provide screen shot

romanofski:
Issue: #90


romanofski:
Dug something out of my spam folder ...

PR #92, opened 2017-10-27
Refactor event handler

romanofski:


romanofski:
Don't merge obviously, this is just a request for comments.

src/UI/App.hs
```diff
@@ -31,8 +36,34 @@ drawUI s =
...
+  handle _ s e = mainEvent s e
+
+dispatch :: EventHandler m => Proxy m -> AppState -> T.BrickEvent Name e -> T.EventM Name (T.Next AppState)
+dispatch m s (T.VtyEvent e) = handle m s e
+
 appEvent :: AppState -> T.BrickEvent Name e -> T.EventM Name (T.Next AppState)
-appEvent s ev = case ev of
+appEvent s (T.VtyEvent ev) =
+  case view asAppMode s of
+    BrowseMail -> handle (Proxy :: Proxy 'BrowseMail) s ev
```
romanofski: Is there a way I can simplify this? I first thought I could write a function which can go from value of type `Mode` to a value of kind `Mode`(???)

src/UI/App.hs
```diff
@@ -31,8 +36,34 @@ drawUI s =
...
+instance EventHandler 'BrowseMail where
+  handle _ s e = handleEvent (view (asConfig . confIndexView . ivKeybindings) s) listEventDefault s e
+
+instance EventHandler 'SearchMail where
+  handle _ s e = handleEventLensed
+
+instance EventHandler 'ManageTags where
+  handle _ s e = mainEvent s e
+
+dispatch :: EventHandler m => Proxy m -> AppState -> T.BrickEvent Name e -> T.EventM Name (T.Next AppState)
```
romanofski: My first idea of a "glue" function between appEvent and the instances...

src/UI/App.hs
```diff
@@ -31,8 +36,34 @@ drawUI s =
         ComposeEditor -> drawComposeEditor s
         Help -> drawHelp s
 
+class EventHandler (m :: Mode) where
+  handle :: Proxy m -> AppState -> Event -> T.EventM Name (T.Next AppState)
+  fallback :: Proxy m -> AppState -> Event -> T.EventM Name (T.Next AppState)
```
romanofski: Not sure if I really need the `fallback` here, since I using the `handleEvent` function would work fine for most items.

src/UI/Index/Main.hs
```diff
@@ -94,12 +94,7 @@ mainEvent s =
 -- Maybe that will remove anything important for the user to see. Do we need
 -- something like an error log?
 listEventDefault :: AppState -> Event -> T.EventM Name (T.Next AppState)
-listEventDefault s e =
-    L.handleListEvent e (view (asMailIndex . miListOfMails) s) >>=
-    \mi' ->
-         M.continue $
-         set (asMailIndex . miListOfMails) mi' s & set asAppMode BrowseMail &
-         set asError Nothing
+listEventDefault s e = M.continue =<< T.handleEventLensed s (asMailIndex . miListOfMails) L.handleListEvent e
```
romanofski: I haven't tested this, but I think the `handleEventLensed` function from [Brick](https://hackage.haskell.org/package/brick-0.20.1/docs/Brick-Types.html#v:handleEventLensed) would be able to do most fallback event handling.

romanofski:
I've been playing with this a bit and I feels the abstraction isn't providing much benefit. It seems what might be better is to abstract it over the targeted widget and the Keybindings we register for that widget than the actual handler. Perhaps I need something which can give me the Keybindings and/or widget for each mode.

PR #95, opened 2017-10-31
Re-factoring the event handler

romanofski:
Use a class to provide lookup of the stored keybindings from the config
and a fallback event handler.

### DIFF
--- a/src/UI/App.hs
+++ b/src/UI/App.hs
@@ -27,7 +27,9 @@ drawUI s =
         SearchMail -> drawMain s
         ManageTags -> drawMain s
         ViewMail -> drawMail s
-        GatherHeaders -> drawInteractiveHeaders s
+        GatherHeadersFrom -> drawInteractiveHeaders s
+        GatherHeadersTo -> drawInteractiveHeaders s
+        GatherHeadersSubject -> drawInteractiveHeaders s
         ComposeEditor -> drawComposeEditor s
         Help -> drawHelp s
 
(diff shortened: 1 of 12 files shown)


src/UI/App.hs
```diff
@@ -32,17 +35,16 @@ drawUI s =
...
-        ComposeEditor -> composeEditor s e
-        Help -> handleHelpEvents s e
-  _ -> M.continue s  -- we only handle Vty events
+appEvent s (T.VtyEvent ev) =
+  case view asAppMode s of
+    BrowseMail -> dispatch (Proxy :: Proxy 'BrowseMail) s ev
+    SearchMail -> dispatch (Proxy :: Proxy 'SearchMail) s ev
+    ManageTags -> dispatch (Proxy :: Proxy 'ManageTags) s ev
+    ViewMail -> dispatch (Proxy :: Proxy 'ViewMail) s ev
+    GatherHeaders -> interactiveGatherHeaders s ev  -- TODO :(
```
romanofski: Major PITA :/ I've fiddled here a bit, but if the current "hardcoded" event handling should be somewhat configurable we'll have to abstract the action also over the Name.

The reason I can not include this into this re-factoring is because `interactiveGatherHeaders` either spawns the editor on Enter or moves the focus on to the next editor widget. Our current action only has a state value, so I won't be able to do much there with the current form of classes (e.g. `Completable`). The only way I could imagine it to work for a keybinding that we have the `Action` abstracted over the name so that I can define an input action for an editor with a specific name and maybe an additonal class which checks that I go from one editor to the next to retrieve `to:`, `from:` and `subject` (or maybe more?) from the user. It all felt a bit over-engineered really.

romanofski: I'm trying out the idea of abstracting the action over the widget name to see where that brings me.

romanofski:
Ok this is now cleaned up, with the rest of the hard coded stuff when composing mail thrown out.

frasertweedale:
@romanofski FYI I probably won't get around to reviewing this until some time in the next week.

romanofski:
Easy. Take your time.

Issue #41, opened 2017-08-20
Test for configuration needed


romanofski:
We need a test which covers the use of a configuration file. Since we settled on a XMonad style config, the test needs to run a customized config with the purebred lib installed. I'm currently not sure how this will work with stack (perhaps not as a user would if a distro package would be installed), but I guess that's something to figure out.

PR #98, opened 2017-11-02
Fixes copy/paste error adding tags

romanofski:
This function is used in the config and is untested (no auto test). I've
realised this happening while using. I'd wish I'd have a better option
at this point to actually test it, but unless we solve issue #41 than
fixing this.


frasertweedale:
obvious ACK.

Issue #96, opened 2017-11-02
Help for focus switch currently not showing which mode it's switching


romanofski:
The help showing keybindings which are performing a focus switch to a different mode are unhelpful:

```
m                                 and then switch focus
```
They're prepended with `and then` and end with `switch focus`. Just doing a `show` on the proxy just appends `Proxy` which isn't very helpful either.

PR #102, opened 2017-11-03
Meaningful description for mode switch

romanofski:
This fixes the description for focus which mode the keybindings are
switching to. Instead of:

  'switch focus'

it now says:

  'switch mode to ManageTags'

issue: #96


frasertweedale:
LGTM

Issue #97, opened 2017-11-02
show mail grouped as threads by default


romanofski:
In order to fully support #35 the first step will be to show mails grouped by threads (which then can be expanded, but that's a separate item).

romanofski:
A bit of status here on how far I'm in. I've tried two approaches:

1. Use one list widget with something like a `NotmuchItem` type, which has two constructors: one which takes `NotmuchMail` type and one which takes a `NotmuchThread` type.
2. Use an additional list widget which holds a `List Name NotmuchThread`.

In regards to (1), it meant more work to distinguish the different types based on their constructor. Some functions we've currently defined for the list which usually holds mails don't make much sense to also apply them to the thread. Not technically since you can set tags on all mails in a thread, but in a UI sense since I fear it's confusing. Also on the rendering side, there was not much benefit with a common type.

Furthermore, even tho there are some similarities to the `Message a n` in notmuch I've ended up realizing they're more different than what they have in common. So all in all it felt like making something work with a really unusual ending.

In regards to (2), is less work, but I need an additional mode to distinguish between browsing mail and browsing threads. Not necessarily bad but I've got trouble generalising actions over an additional list holding a different type. Furthermore, I thought about supporting separate modes: one mode showing all mails in the mailbox as a flat list and one, which gives you a thread view (from notmuch) -> showing a flat list of mails per thread. That works fine, yet to support both modes independently I need a way to distinguish that you're not quitting the application from the mail-per-thread-view, but rather go back to showing threads.

PR #103, opened 2017-11-04
Show threads by default

romanofski:
By default purebred now shows only top level messages

Issue: #97

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -15,8 +15,8 @@ import qualified Data.Vector as Vec
 import System.Process (readProcess)
 import qualified Data.Text as T
 import Data.Text.Encoding (decodeUtf8, encodeUtf8)
-import Types (NotmuchMail(..), NotmuchSettings, nmDatabase, mailId, mailTags)
-import Control.Lens (view, over, set)
+import Types
+import Control.Lens (view, over, set, firstOf, folded)
 
 import Notmuch
 import Notmuch.Search
@@ -93,12 +93,12 @@ messageToMail
 messageToMail m = do
     tgs <- tags m
     let tgs' = decodeUtf8 . getTag <$> tgs
-    NotmuchMail <$>
-      (decodeUtf8 . fromMaybe "" <$> messageHeader "Subject" m) <*>
-      (decodeUtf8 . fromMaybe "" <$> messageHeader "From" m) <*>
-      messageDate m <*>
-      pure tgs' <*>
-      messageId m
+    NotmuchMail
+      <$> (decodeUtf8 . fromMaybe "" <$> messageHeader "Subject" m)
+      <*> (decodeUtf8 . fromMaybe "" <$> messageHeader "From" m)
+      <*> messageDate m
+      <*> pure tgs'
+      <*> messageId m
 
 getDatabasePath :: IO FilePath
 getDatabasePath = getFromNotmuchConfig "database.path"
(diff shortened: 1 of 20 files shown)


romanofski:
Corresponding change is in purebred-mua/hs-notmuch#17

src/Storage/Notmuch.hs
```diff
@@ -40,6 +40,22 @@ getMessages s settings =
...
+getTopMessages
+  :: (HasThreads Query, MonadError Error m, MonadIO m)
+  => T.Text
+  -> NotmuchSettings FilePath
+  -> m (Vec.Vector NotmuchMail)
+getTopMessages s settings =
+  bracketT (databaseOpenReadOnly (view nmDatabase settings)) databaseDestroy go
+  where
+    go db = do
+        tds <- query db (FreeForm $ T.unpack s) >>= threads >>= mapM threadToplevelMessages
```
frasertweedale: Instead of `mapM` you can often use `traverse` which uses the weaker `Applicative` abstraction.
I think you probably can use `traverse` in this case.

src/Storage/Notmuch.hs
```diff
@@ -40,6 +40,22 @@ getMessages s settings =
...
+        let flattened = foldl (++) [] tds
```
frasertweedale: Use `(<>)` instead of `(++)`, and I think you probably want `foldr` instead of `foldl`.

Or because it is `[[a]]` you could replace `flattened` with `Control.Monad.join`.

src/Storage/Notmuch.hs
```diff
@@ -40,6 +40,22 @@ getMessages s settings =
...
+        msgs <- liftIO $ mapM messageToMail flattened
```
frasertweedale: `traverse`

src/UI/Actions.hs
```diff
@@ -301,7 +301,7 @@ reloadMails = Action "reload list of mails" applySearch
 --
 applySearch :: AppState -> T.EventM Name AppState
 applySearch s =
-   runExceptT (Notmuch.getMessages searchterms (view (asConfig . confNotmuch) s))
+   runExceptT (Notmuch.getTopMessages searchterms (view (asConfig . confNotmuch) s))
```
frasertweedale: I'm unsure about whether we really want this to be the default or not.  I can forsee a need to make it configurable in the future.  But I'm ok with the change for now.  I'm not even using purebred in anger yet... you are (probably) the only person in the world who is right now, so I'll defer to your judgement :)

romanofski: Yeah I'm actually not sure about this either. Most clients follow the approach of showing the conversation. That is: thread -> unread mails expanded with folded quoting. I grew up with mutt showing individual mails. Thinking about it however I can't think of a case where I would want to use a thread list with only one mail expanded (mutt style). Do you have a good one?

frasertweedale: Nope.  Go with your gut, I'll go along with it :)

src/Storage/Notmuch.hs
romanofski: Good to know that instead of `mapM` rather use `traverse`... changed this.

src/Storage/Notmuch.hs
romanofski: Forgetting this all the time, but the code has changed since then.

romanofski:
Alright I'm sick of it. If you have some nits to do this better (esp. I'm wondering if the `selectedItemHelper` can be done better... maybe a class?) I'm all ears, otherwise I need the other PR merged so I can rebase on top of it and then merge it as well to pick something else.

src/Storage/Notmuch.hs
```diff
@@ -24,6 +24,27 @@ import Notmuch.Util (bracketT)
 
 import Error
 
+class ManageTags a where
+  setTags :: a -> [T.Text] -> a
+  addTags :: a -> [T.Text] -> a
+  removeTags :: a -> [T.Text] -> a
+  getTags :: a -> [T.Text]
```
frasertweedale: replace all of this with a lens.  Then if you want the convenience functions define them at the top level.  example:
```haskell
class HasTags a where
  tags :: Lens' a [T.Text]
  writeTags :: (MonadError e m, MonadIO m, AsError e) => FilePath -> a -> m a

setTags, addTags, removeTags :: (HasTags a) => [T.Text] -> a -> a
...
```

src/UI/Actions.hs
```diff
@@ -285,18 +285,24 @@ setUnread =
     , _aAction = (liftIO . updateReadState Notmuch.addTags)
     }
 
-mailIndexUp :: Action 'BrowseMail AppState
-mailIndexUp =
+listUp :: Action m AppState
```
frasertweedale: Should we implement a class for *"mode with a navigable list"* and define it in terms of
that class?

src/Storage/Notmuch.hs
romanofski: Great! Why have I not thought of that before. Cheers!

src/Storage/Notmuch.hs
```diff
@@ -10,12 +10,12 @@ import Control.Monad.Except (MonadError, throwError)
 import qualified Data.ByteString as B
 import Data.Traversable (traverse)
 import Data.List (union, notElem)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, listToMaybe)
```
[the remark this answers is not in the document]
romanofski: oh! cool! Cheers.

src/UI/Actions.hs
romanofski: Yes I'd like this idea. Created #112 for this.

romanofski:
Last push with all the changes incorporated. If that runs through I'll merge ... tomorrow.

Issue #112, opened 2017-12-01
Mode with navigatable list


romanofski:
From #103 @frasertweedale suggested:

> Should we implement a class for "mode with a navigatable list" and define it in terms of
that class? Because at the moment you can use listUp with any mode, including those that don't have a list...

This is a backlog item to implement a class for "mode with navigatable list.

romanofski:
Fixed with #266

PR #116, opened 2017-12-06
Select next unread

romanofski:


src/UI/Actions.hs
```diff
@@ -388,8 +388,22 @@ removeTags ts =
...
+                                 fx m = Notmuch.isNew (view (asConfig . confNotmuch . nmNewTag) s) m
+                                 next = nextUnread vec cur fx
+                             in pure $ over (asMailIndex . miListOfMails) (L.listMoveTo (fromJust next)) s)
+         }
+
 -- Function definitions for actions
 --
+-- | Return the next unread item or default to the currently selected.
+nextUnread :: Vector.Vector a -> Maybe Int -> (a -> Bool) -> Maybe Int
+nextUnread vec cur fx = Vector.findIndex fx (Vector.drop ((fromJust cur) + 1) vec) <|> cur
```
romanofski: Not very happy about the `fromJust` use here since it could lead to an Exception. However if I simply pattern match the `cur` it should be fine? Perhaps use the `maybe` function?

src/UI/Actions.hs
```diff
@@ -388,8 +388,22 @@ removeTags ts =
 reloadList :: Action 'BrowseThreads AppState
 reloadList = Action "reload list of threads" applySearch
 
+selectNextUnread :: Action 'BrowseMail AppState
```
romanofski: I've currently only implemented it for `'BrowseMail` and I think that's enough for now. However I can see that we want to change it once we have #112 implemented.

src/Storage/Notmuch.hs
```diff
@@ -40,6 +40,9 @@ removeTags tgs = over tags (filter (`notElem` tgs))
 getTags :: (ManageTags a) => a -> [T.Text]
 getTags = view tags
 
+isNew :: (ManageTags a) => T.Text -> a -> Bool
+isNew ignoredTag x = ignoredTag `elem` (view tags x)
```
frasertweedale: `ignoredTag` seems like a strange choice for the arg name.  Even the name of the function
is a bit off - seems like ``hasTag t x = t `elem` view tags x`` would be a better names.

src/UI/Actions.hs
```diff
@@ -388,8 +388,22 @@ removeTags ts =
...
+selectNextUnread =
+  Action { _aDescription = "select next unread"
+         , _aAction = (\s -> let vec = (view (asMailIndex . miListOfMails . L.listElementsL) s)
+                                 cur = (view (asMailIndex . miListOfMails . L.listSelectedL) s) <|> Just 0
```
frasertweedale: This line can be `cur = vec <|> Just 0`

src/UI/Actions.hs
```diff
@@ -388,8 +388,22 @@ removeTags ts =
...
+                                 fx m = Notmuch.isNew (view (asConfig . confNotmuch . nmNewTag) s) m
+                                 next = nextUnread vec cur fx
+                             in pure $ over (asMailIndex . miListOfMails) (L.listMoveTo (fromJust next)) s)
```
frasertweedale: `fromJust` is non-total; don't use it.

src/UI/Actions.hs
frasertweedale: Again, `fromJust` is unsafe.

frasertweedale: Yeah `maybe` or `fromMaybe`.

src/Storage/Notmuch.hs
romanofski: Yeah I think I stuffed that one up. Sorry :(

src/UI/Actions.hs
romanofski: Could it be a mistake? One returns the vector the other view the selected list element ...

romanofski:
Ugh I'm amazed how often we run into build errors which are PPA, network or travis related ...

src/UI/Actions.hs
frasertweedale: whoops, I misread.  pls ignore.

src/UI/Actions.hs
```diff
@@ -388,8 +390,22 @@ removeTags ts =
...
+                                 fx m = Notmuch.hasTag (view (asConfig . confNotmuch . nmNewTag) s) m
+                                 next = nextUnread vec cur fx
+                             in pure $ over (asMailIndex . miListOfMails) (L.listMoveTo (fromMaybe 0 next)) s)
+         }
+
 -- Function definitions for actions
 --
+-- | Return the next unread item or default to the currently selected.
+nextUnread :: Vector.Vector a -> Maybe Int -> (a -> Bool) -> Maybe Int
+nextUnread vec cur fx = Vector.findIndex fx (Vector.drop ((fromMaybe 0 cur) + 1) vec) <|> cur
```
frasertweedale: Seems like you should be adding the current offset to the result of `findIndex`?

frasertweedale:
```haskell
findIndexWithOffset i f = fmap (i+) . V.findIndex f . V.drop i

seekIndex i f = fromMaybe i . findIndexWithOffset (i + 1) f
```

romanofski: Alright got this in. I'll try to find a property to maybe test this.

romanofski:
Hmpf.. there is one test failing which I can't reproduce anywhere... not even on a Ubuntu box with the same version...

frasertweedale:
@romanofski looks escape-code related.  I'll try to have a look at it tonight.

frasertweedale:
Quite possibly caused by a different value of `TERM` env var or `terminfo` available in the CI env.
Can we update CI to log some of this stuff?

frasertweedale:
So the problem is a combination of:

- test environment is slow
- the string we were check for to indciate that the change has occurred was already present on the screen

So the test assumed everything had been redrawn and moved on to the next test.  But everything
had not been redrawn yet.  Anyhow Ima merge this now :)

frasertweedale:
Merged (4a36433).

romanofski:
Many thanks @frasertweedale

Issue #108, opened 2017-11-25
Sometimes tmux fails to start


romanofski:
When running the tests, tmux sometimes fails to start:

```
Exception: callProcess: tmux "new-session" "-x" "80" "-y" "24" "-d" "-s" "purebredtest" "-n" "purebred" (exit 1): failed
```
It would be good in that case to show more information about the environment in order to debug the problem.

PR #117, opened 2017-12-06
Show IOError in case of tmux setup failure

romanofski:
Closes issue: #108


romanofski:
Could probably be merged even if the tests for some of the pull requests don't fail.

PR #122, opened 2017-12-11
Improve "Item <m> of <n>" UI

frasertweedale:
Currently we display "Item <index> of <n>" e.g. "Item 0 of 1".
Users probably want 1-based indexing, especially so that it can
align at the end of the list, i.e. "Item 7 of 7".  Also improve the
output when there are no items by saying "No items" instead of "Item
0 of 0".


romanofski:
Oof:
```
Exception during setUp: callProcess: tmux "new-session" "-x" "80" "-y" "24" "-d" "-s" "purebredtest" "-n" "purebred" (exit 1): failed

failed to connect to server

failed to connect to server
```
Might be interesting to cat the tmux server log output to check if the server dies during test runs.

romanofski:
Happy to merge unless you want to fiddle with the builds (doubt it).

PR #128, opened 2017-12-13
selectNextUnread: if no current item return offset 0

frasertweedale:
```
a772d2e (Fraser Tweedale, 17 minutes ago)
   selectNextUnread: if no current item return offset 0

   Only apply `findIndexWithOffset` if there is a current selection. Otherwise
   the first item should be selected.

   Fixes: https://github.com/purebred-mua/purebred/issues/126

55eb569 (Fraser Tweedale, 21 minutes ago)
   reformat selectNextUnread action

   Long lines are long.  Reformat.
```

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -399,14 +399,13 @@ selectNextUnread =
          , _aAction = \s ->
            let
              vec = view (asMailIndex . miListOfMails . L.listElementsL) s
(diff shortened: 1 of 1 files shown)


frasertweedale:
@romanofski please check the behaviour of this.  I have not run it :)

romanofski:
Yes is worky. 👍

Issue #130, opened 2017-12-13
create `withDatabase` helper function


frasertweedale:
We have a bunch of uses of `bracketT` that just open a database, do the thing, then close
the database.  We should extract a helper function for this use case.

In `src/Storage/Notmuch.hs`:
```
64:  bracketT (Notmuch.databaseOpenReadOnly (view nmDatabase settings)) Notmuch.databaseDestroy go
74:  bracketT (Notmuch.databaseOpenReadOnly dbpath) Notmuch.databaseDestroy go
85:  bracketT (Notmuch.databaseOpen dbpath) Notmuch.databaseDestroy (tagsToMessage nmtags (view mailId m))
96:  void $ bracketT (Notmuch.databaseOpen dbpath) Notmuch.databaseDestroy (go tgs mgs)
149:  bracketT (Notmuch.databaseOpenReadOnly (view nmDatabase settings)) Notmuch.databaseDestroy go
164:  bracketT (Notmuch.databaseOpenReadOnly fp) Notmuch.databaseDestroy go
```

alexeyzab:
Hi! I'd like to add this.

romanofski:
Hey @alexeyzab ! Welcome! Assign and hack away :)

frasertweedale:
@alexeyzab welcome!  Your contribution will be most appreciated.

romanofski:
Closing this. Thanks to the contribution from @alexeyzab :-)

PR #132, opened 2017-12-14
Add a `withDatabase` helper

alexeyzab:
Related to #130.

Note that, in its current form, this function specifies
a `Notmuch.RW` (read/write) database access, whereas previously
we were using `Notmuch.databaseOpenReadOnly` in most of the
functions.

The reason for that is the fact that `hs-notmuch` does not
currently export the `Mode a` typeclass, which would allow
us to parametrize our helper function by using it as a
constraint. While `databaseOpen` and `databaseOpenReadOnly`
are identical, we are losing a type level constraint here.

There is probably a better way of addressing it here. Any tips are appreciated!

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -6,7 +6,7 @@
 module Storage.Notmuch where
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
-import Control.Monad.Except (void, MonadError, throwError)
+import Control.Monad.Except (void, MonadError, throwError, ExceptT)
 import qualified Data.ByteString as B
 import Data.Traversable (traverse)
 import Data.List (union, notElem)
@@ -51,6 +51,23 @@ instance ManageTags NotmuchThread where
   tags = thTags
   writeTags = setNotmuchThreadTags
 
+-- | A helper function for opening, performing work,
+-- and closing the database
+withDatabase
+  :: (Notmuch.AsNotmuchError e, Notmuch.Mode a, MonadError e m, MonadIO m)
+  => FilePath
+  -> (Notmuch.Database a -> ExceptT e IO c)
+  -> m c
+withDatabase dbpath = bracketT (Notmuch.databaseOpen dbpath) Notmuch.databaseDestroy
+
+-- | Same as 'withDatabase', but the database connection
+-- is read-only
+withDatabaseReadOnly
+  :: (Notmuch.AsNotmuchError e, MonadError e m, MonadIO m)
+  => FilePath
+  -> (Notmuch.Database Notmuch.RO -> ExceptT e IO c)
+  -> m c
+withDatabaseReadOnly = withDatabase
 
 -- | creates a vector of parsed mails from a not much search
 -- Note, that at this point in time only free form searches are supported. Also,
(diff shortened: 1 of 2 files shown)


frasertweedale:
@alexeyzab yeah, hold that thought; I'll update hs-notmuch to export `Mode`, and then we can implement the desired function.  (Later today...)

frasertweedale:
@alexeyzab here ya go: https://github.com/purebred-mua/hs-notmuch/commit/75150384a5815d2ebabfe7bb45151e2be39cc90e.
Update the hs-notmuch commit hash in `stack.yaml` as part of your change.

alexeyzab:
@frasertweedale I ended up rewriting the functions a bit to have the long type signature on a separate line by using `let in`. If that's undesirable, I can rewrite it back.

src/Storage/Notmuch.hs
```diff
@@ -51,28 +52,38 @@ instance ManageTags NotmuchThread where
...
 getMessages
-  :: (MonadError Error m, MonadIO m)
+  :: forall m. (MonadError Error m, MonadIO m)
   => T.Text
   -> NotmuchSettings FilePath
   -> m (Vec.Vector NotmuchMail)
 getMessages s settings =
-  bracketT (Notmuch.databaseOpenReadOnly (view nmDatabase settings)) Notmuch.databaseDestroy go
+  let withDb = withDatabase (view nmDatabase settings) :: (Notmuch.Database Notmuch.RO -> ExceptT Error IO c) -> m c
+  in withDb go
```
frasertweedale: Instead of using `ScopedTypeVariables` and these type signatures which are... a bit heavy, let's create a variant
of `withDatabase` specialised to `RO` database.  So:
```haskell
withDatabase
  :: (Notmuch.AsNotmuchError e, Notmuch.Mode a, MonadError e m, MonadIO m)
  => FilePath
  -> (Notmuch.Database a -> ExceptT e IO c)
  -> m c
withDatabase dbpath = bracketT (Notmuch.databaseOpen dbpath) Notmuch.databaseDestroy

withDatabaseReadOnly
  :: (Notmuch.AsNotmuchError e, MonadError e m, MonadIO m)
  => FilePath
  -> (Notmuch.Database Notmuch.RO -> ExceptT e IO c)
  -> m c
withDatabaseReadOnly = withDatabase
```

Then we can just use `databaseOpenReadOnly` where appropriate, and avoid `ScopedTypeVariables`.

alexeyzab: Thanks for pointing this out!

romanofski:
sigh
```
W: The repository 'http://ppa.launchpad.net/couchdb/stable/ubuntu trusty Release' does not have a Release file.

W: Failed to fetch http://ppa.launchpad.net/chris-lea/redis-server/ubuntu/dists/trusty/main/binary-amd64/Packages  Unable to connect to ppa.launchpad.net:http:
```

romanofski:
Btw. @frasertweedale with contributions coming in I'd like to keep track of the contributors. Shall we create an `AUTHORS` file?

frasertweedale:
Pushed to master: https://github.com/purebred-mua/purebred/commit/8560adae8db3e2107dec073b63ebd8b6a5b0e357.

frasertweedale:
Thanks @alexeyzab!

@romanofski I reckon `git shortlog -esn` does a decent job of listing contributors.  I am not averse to an `AUTHORS` or `CONTRIBUTORS` file but it is one more thing to maintain whereas git already tell us this info...

PR #135, opened 2017-12-30
Clear errors with the next key handler

romanofski:
This sets the error to nothing with the next key press event. This has
the potential of errors being missed if the key presses from the user
happen to quickly, but anything more elaborate is not of priority now.

Fixes https://github.com/purebred-mua/purebred/issues/73

### DIFF
--- a/src/UI/Keybindings.hs
+++ b/src/UI/Keybindings.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE KindSignatures #-}
 module UI.Keybindings where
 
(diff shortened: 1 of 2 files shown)


frasertweedale:
I'm getting test failures with this now.  Seems it either regressed somehow or the way we are testing it may be wrong?

romanofski:
Hm matching against `openBinaryFile` fixes it for me (due to changes in the code against refactor/purebred-email):
```Haskell
    liftIO $ step "shows error message"
    sendKeys "Enter" (Literal "FileReadError")
      >>= assertSubstrInOutput "openBinaryFile: does not exist"
```
and rerunning with:
```
$ stack test --test-arguments='-p "tests/user*/error*"'
```
Am I missing something else?

frasertweedale:
I'm experiencing the problems on `master`.  The error is not being cleared.

frasertweedale:
Aah, I just forgot to `stack install`.  Never mind :)

romanofski:
Yeah possible pitfall. Sorry. One of the reasons why I thought about moving those acceptance tests into a separate package.

PR #136, opened 2017-12-30
use purebred-email for parsing mail

frasertweedale:
This patchset switches mail parsing to purebred-email.
Please try it out and see where the glaring problems are.

One known gap: content-transfer-encoding is not automatically performed.
Neither do we do it in purebred itself.  So bodies using quoted-printable or
base64 will be displayed as-is.  I will try to address this in the coming days.

If you notice any other epic fails please let me know :)

### DIFF
--- a/src/Storage/ParsedMail.hs
+++ b/src/Storage/ParsedMail.hs
@@ -9,39 +9,53 @@ import Codec.MIME.Parse (parseMIMEMessage)
 import Codec.MIME.Type
        (MIMEParam(..), MIMEValue(..), mime_val_headers)
 import Control.Exception (try)
+import Control.Lens (firstOf)
 import Control.Monad.Except (MonadError, throwError)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import qualified Data.ByteString as B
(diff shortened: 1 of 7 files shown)


romanofski:
ooooh its happening!!! 🥇

src/Storage/ParsedMail.hs
```diff
@@ -16,46 +13,33 @@ import qualified Data.ByteString as B
...
-getTo :: ParsedMail -> T.Text
-getTo (MIMEMail v) = findHeader v "to"
-getTo (PurebredEmail msg) = getHeader "to" msg
+getSubject :: Message MIME -> T.Text
+getSubject = getHeader "subject"
 
-findHeader :: MIMEValue -> T.Text -> T.Text
-findHeader m name = T.strip . paramValue . head $ filter (\x -> paramName x == name) $ mime_val_headers m
+getTo :: Message MIME -> T.Text
+getTo = getHeader "subject"
```
romanofski: mistake? shouldn't it be "to"?

frasertweedale: yes it should.  thanks!

PR #138, opened 2017-12-31
Tests/run hlint

romanofski:


romanofski:
Hm... now realising that just running wholesale is not applicable. Tempted to close the pull request without merging this (or maybe just the hlint suggestions) instead of conditionally running hlint. Let me know what you think.

frasertweedale:
@romanofski yeah let's just merge the cleanups suggested.  I don't think we need to run HLint as part of build/test process.

romanofski:
Yeah alright, dropped the travis change.

PR #140, opened 2018-01-03
assorted ci, readme and .cabal improvements

frasertweedale:


.travis.yml
```diff
@@ -17,33 +17,33 @@ before_install:
   - pushd /tmp/notmuch && ./configure --prefix=/usr --without-emacs --without-ruby && make && sudo make install && popd
 
 script:
-  - stack $STACK_ARGS --system-ghc --no-terminal --skip-ghc-check install
-  - DEBUG=1 stack $STACK_ARGS --system-ghc --no-terminal --skip-ghc-check test 2> /tmp/test.log
+  - stack $STACK_ARGS --no-terminal install
```
romanofski: Cool if this works. Now seeing this change, I actually don't even know why I used the ppa.

frasertweedale: The ppa is useful if you're using plain old cabal.  In fact I intend to add this into the test matrix at some point, because it will make testing with development versions of GHC easier (hvr's ppa adopts prerelease builds much more eagerly than Stackage nightly does).  But that's a task for another day.

PR #143, opened 2018-01-04
tests: use "keepalive" tmux session and unique session names

frasertweedale:
Let's hammer Travis with this commit and see if it looks like it
resolves the flaky tests.

### DIFF
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -8,6 +8,7 @@ module TestUserAcceptance where
 
 import qualified Data.Text as T
 import System.IO.Temp (createTempDirectory, getCanonicalTemporaryDirectory)
+import Data.Functor (($>))
 import Data.Ini (parseIni, writeIniFileWith, KeySeparator(..), WriteIniSettings(..))
 import Data.Semigroup ((<>))
 import Control.Concurrent (threadDelay)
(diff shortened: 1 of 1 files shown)


test/TestUserAcceptance.hs
```diff
@@ -282,16 +279,35 @@ assertRegex regex out = liftIO $ assertBool
...
 
-setUp :: IO (String, String, String)
-setUp = do
-  let sessionname = defaultSessionName
-  (testdir, testmdir) <- setUpTmuxSession sessionname >> setUpTempMaildir
-  pure (testdir, testmdir, sessionname)
+-- | Set up a test session.
+setUp :: Int -> IO Env
+setUp i = do
+  let sessionName = defaultSessionName <> show i
```
romanofski: That would create sessions like `purebredtest1,2,3,4...` right. You do have the test case name, why not use that as an additional identifier? That'll help when debugging any failures.

test/TestUserAcceptance.hs
```diff
@@ -37,20 +38,24 @@ data Condition
...
-        , testCanToggleHeaders
-        , testSetsMailToRead
-        , testErrorHandling
-        , testHelp
-        , testManageTagsOnMails
-        , testManageTagsOnThreads
-        ]
-
-testManageTagsOnMails :: TestTree
+  withResource pre post $ \_ ->
```
romanofski: heh ... funny how "easy" that is to just add group specific setup and teardown functions ...

test/TestUserAcceptance.hs
frasertweedale: I'll do this in a separate commit.

romanofski: Fair enough

frasertweedale: PR: https://github.com/purebred-mua/purebred/pull/145

test/TestUserAcceptance.hs
frasertweedale: Yeah I was pleased with ease of discovery and ease of use of the interface.

PR #146, opened 2018-01-08
switch tag parsing to attoparsec, and unrelated cleanups

frasertweedale:
```
3695c4b (Fraser Tweedale, 2 minutes ago)
   avoid non-total function decodeUtf8

   Data.Text.Encoding.decodeUtf8 is a non-total function which throws an
   exception on invalid input.  Change to 'decodeLenient' which replaces
   invalid code points with a placeholder character.

57b7f33 (Fraser Tweedale, 4 minutes ago)
   remove unused dependencies

44d2f99 (Fraser Tweedale, 14 minutes ago)
   use quickcheck-instances instead of quickcheck-text

   The more popular quickcheck-instances package provides orphan Arbitrary
   instances not only for 'text' but many other common library.  Switch to it,
   and remove 'time'-related orphan instances.

74ed5be (Fraser Tweedale, 36 minutes ago)
   switch tag parsing back to attoparsec

   We want to avoid depending on two different parser libraries. So switch
   back to attoparsec for parsing tags.

   This refactor also changes Purebred to use the Notmuch.Tag type internally.
    User input is parsed into a Notmuch.Tag during parsing. Anything that will
   not be accepted by libnotmuch is rejected at that point, so there is now
   one less error that can occur later on when trying to write tags.

   attoparsec's error handling is not as nice as parsec so add a couple of
   helper parsers for getting nicer errors out of it.  The error message did
   change a little, so update the test accordingly.

   An orphan Arbitrary instance is added for Notmuch.Tag.  The Tag is of a
   random length (between 1 and the max length), and consists of random
   printable ASCII characters excluding SPACE.  Bump the hs-notmuch dependency
   to a commit where 'tagMaxLen' is exported (we need it for the Arbitrary
   instance).
```

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -110,7 +110,7 @@ test-suite unittests
                      , brick
                      , tasty-hunit
                      , tasty-quickcheck
-                     , quickcheck-text
+                     , quickcheck-instances
                      , tasty
                      , directory
                      , process
@@ -121,7 +121,6 @@ test-suite unittests
                      , temporary
                      , text
                      , resourcet
-                     , time
                      , regex-posix
                      , mtl
                      , lens
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -73,7 +73,6 @@ library
                      , text
                      , process
                      , mime-mail
-                     , network
                      , directory
                      , bytestring
                      , time
@@ -81,7 +80,6 @@ library
                      , optparse-applicative >= 0.13
                      , xdg-basedir
                      , filepath
-                     , unix
                      , mtl
                      , exceptions
                      , purebred-email
@@ -107,20 +105,16 @@ test-suite unittests
   default-language:    Haskell2010
   build-depends:       base
                      , purebred
-                     , brick
+                     , tasty
                      , tasty-hunit
                      , tasty-quickcheck
                      , quickcheck-instances
-                     , tasty
                      , directory
                      , process
-                     , unbounded-delays
                      , bytestring
-                     , network
                      , ini
                      , temporary
                      , text
-                     , resourcet
                      , regex-posix
                      , mtl
                      , lens
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -85,7 +85,7 @@ library
                      , mtl
                      , exceptions
                      , purebred-email
-                     , parsec
+                     , attoparsec
 
 executable purebred
   hs-source-dirs:      app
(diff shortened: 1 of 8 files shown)


purebred.cabal
```diff
@@ -73,15 +73,13 @@ library
...
                      , mime-mail
-                     , network
                      , directory
                      , bytestring
                      , time
                      , case-insensitive
                      , optparse-applicative >= 0.13
                      , xdg-basedir
                      , filepath
-                     , unix
```
romanofski: Did you use anything to find these?

src/Storage/Notmuch.hs
```diff
@@ -60,21 +61,21 @@ applyTagOp (RemoveTag t) = removeTags [t]
 applyTagOp ResetTags = set tags []
 
 class ManageTags a  where
-    tags :: Lens' a [T.Text]
+    tags :: Lens' a [Tag]
 
-setTags :: (ManageTags a) => [T.Text] -> a -> a
+setTags :: (ManageTags a) => [Tag] -> a -> a
```
romanofski: Aaah of course. Yes... really good idea to change it straight into a `Notmuch.Tag`. Cheers!

purebred.cabal
frasertweedale: There exists https://hackage.haskell.org/package/packunused but on this occasion I just
tried removing each of the deps that I wasn't sure were used, and if it still built, well...  :)

romanofski: Ah right... cheers.

test/TestMail.hs
```diff
@@ -1,16 +1,19 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
romanofski: Do we still need this? Maybe my misunderstanding, but I thought the switch to quickcheck-instances would have eliminated the need?

romanofski:
Good refactor. I like especially the changes switching back to attoparsec. Something to learn from ...

test/TestMail.hs
frasertweedale: We now have the orphan instance for `Notmuch.Tag`

frasertweedale:
@romanofski thanks for review and approval.  As you can see, you can still get useful errors out of attoparsec, but you have to do a bit of extra stuff and they are still not quite as rich as parsec.  But parsec is kinda overkill here anyway - it's aimed at source code and other larger, more complex text files.

PR #147, opened 2018-01-12
Feature file browser

romanofski:
This is adding just the file browser which allows browsing the local file system in a very basic way.

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -59,6 +59,7 @@ library
                      , UI.Help.Keybindings
                      , UI.Utils
                      , Purebred.Tags
+                     , Purebred.System.Directory
                      , UI.FileBrowser.Main
                      , UI.FileBrowser.Keybindings
                      , Config.Main
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -59,6 +59,8 @@ library
                      , UI.Help.Keybindings
                      , UI.Utils
                      , Purebred.Tags
+                     , UI.FileBrowser.Main
+                     , UI.FileBrowser.Keybindings
                      , Config.Main
                      , Storage.Notmuch
                      , Storage.ParsedMail
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -10,6 +10,7 @@ import qualified Brick.Widgets.Edit as E
 import qualified Graphics.Vty as V
 import System.Environment (lookupEnv)
 import Data.Maybe (fromMaybe)
+import UI.FileBrowser.Keybindings (fileBrowserKeybindings)
 import UI.ComposeEditor.Keybindings (composeEditorKeybindings)
 import UI.GatherHeaders.Keybindings
        (interactiveGatherHeadersKeybindings,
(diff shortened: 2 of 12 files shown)


romanofski:
This is still a bit work in progress, especially from what I can see is:

1) add acceptance test(s)
2) the file browser is not wired up to the mail composition so therefore doesn't do anything yet
3) some state names are a bit "off"

Yet I thought I share it, because I have questions for the last two TODO items:
For 2) The current key binding is defined as:
```Haskell
 Keybinding (V.EvKey V.KEnter []) (changeDirectory `chain` continue)
```
Reason being, I want to enter directories, but add files. Which means just for adding attachments, I could hard code the behaviour into the `changeDirectory` action. Yet if we want to re-use the file browser for e.g. saving attachments I'd need something different. So, with the current architecture the only way I can see this accomplished is to encode this behaviour in the `ModeTransition (s :: Mode) (d :: Mode)` class, because that's where both use cases differ. Any thoughts here?

As to what to do with the creating attachments I'm already clear, since I have code which initialises the `Compose` with an empty [Mail](https://hackage.haskell.org/package/mime-mail-0.4.11/docs/Network-Mail-Mime.html#t:Mail) instead of a temporary file. So grabbing the toggled files from the file browser and converting them to attachments should be straight forward (tm).

For 3) I'm currently keeping a persistent search path or file browsing path in two places. One in the configuration which acts as the initial path. When the user switches to the `AddAttachment` mode, the configuration path is copied into the file browser and then modified in place with lenses. The configuration file path is kept as it is. This feels a bit odd (the copying and modifying from the config) and I'm interested if you have perhaps a better idea.

romanofski:
Ah and a PS: I was thinking of maybe squashing all the commits into one. So let me know what you prefer.

frasertweedale:
Don't squash.  Some of the commits do different things.  COmmits are cheap mate.
Will try to review this w/e.

romanofski:
I recorded an ascii session which uses the file browser diving in to directories and going up the file system hierarchy. Maybe that helps with the reviewing.

[asciicast](https://asciinema.org/a/dPnWdQ3Ouos2pm24KetUzIdR1)

romanofski:
I'm just realising that my initial idea for 2) to use a function in the `ModeTransition` class feels like a hack. The keybinding as outlined above is currently:

```Haskell
 Keybinding (V.EvKey V.KEnter []) (changeDirectory `chain` continue)
```

Using the `ModeTransition` changes this to:

```Haskell
 Keybinding (V.EvKey V.KEnter []) (changeDirectory `chain'` (focus :: Action 'ComposeEditor AppState) `chain` continue)
```
which means that the function in the class does two things:
a) check if we've selected a file
b) if no file is selected, but a directory, don't change the mode but carry on

A similar thing could probably accomplished in a more explicit `Action`. I'm still wondering if we need conditions in the keybindings ...
```

romanofski:
I'm closing this pull request for now, because
a) I've found a way to work with the implemented file browser
b) having trouble with the overall usability of the composing editor, mainly in regards to focus and stealing keybindings.

I'm currently looking for better ways to get this view to work without a exploding number of modes and their widgets. Perhaps the `Form` implementation might help, but I'm also thinking about the possibility of refactoring the modes a bit. Hence, doing some research.

PR #149, opened 2018-01-13
Updated RPM package

romanofski:


romanofski:
Don't merge yet. I've noticed a problem with this spec file update. The changes I've made to build a proper package are not included here. Update follows.

romanofski:
Alright... now I got the right set.

purebred.spec
```diff
@@ -80,29 +99,41 @@ This package provides the Haskell %{pkg_name} library development files.
...
 
 
-%files devel -f %{name}-devel.files
-%doc README.html README.md
-%{_bindir}/%{pkg_name}
+%files -n ghc-%{name}-devel -f ghc-%{name}-devel.files
+%doc README.md
 
+%files config
+%{_datadir}/purebred/config.hs
```
frasertweedale: Is this the default config?  If so wouldn't it be better to ship as part of base package?

purebred.spec
```diff
@@ -19,58 +19,77 @@ BuildRequires:  ghc-Cabal-devel
...
 Requires:       ghc-compiler = %{ghc_version}
 Requires(post): ghc-compiler = %{ghc_version}
 Requires(postun): ghc-compiler = %{ghc_version}
-Requires:       %{name}%{?_isa} = %{version}-%{release}
+Requires:       ghc-%{name}%{?_isa} = %{version}-%{release}
+
+%description -n ghc-%{name}-devel
+This package provides the Haskell %{name} library development files.
+
+%package basic
```
frasertweedale: I would have thought `purebred-base` is a more idiomatic package name.
But could we even make it just `purebred`?

romanofski: The 'user installable' package is called purebred. The library development files are shipped as ghc-purebred-devel. I want to stay consistent with the [packaging guidelines](https://fedoraproject.org/wiki/Packaging:Haskell#Package_Naming) which, for libraries suggest to prefix with `ghc-`.

romanofski: In regards to `basic` vs `base` ... no hard feelings and I'm happy to rename it to `base`. Actually sounds better ...

romanofski: Ah.. one more thing: I picked `basic` because I looked at the xmonad package. I don't want to repeat their mistakes, but that's the reason I had chosen `basic`.

purebred.spec
romanofski: Again... peeked at the xmonad package as an example. I suppose you can still install `purebred-base` without the config and operate the client. Especially if the default key bindings are all you need. No hard feelings here either, tho arguably we could ship different config flavours for key bindings (e.g. vim-like, mutt-like, etc).

frasertweedale: overkill IMO.  but i guess the option is there.

frasertweedale:
Anyhow mate, RPM packaging is you're department so if you're happy with this just push it :)

romanofski:
Yeah @frasertweedale I agree at this point we don't really need it. Am I in the moot to change it back? Not really. I expect changes to it if we want to actually file a package review to get the package into Fedora. So for now I leave it in a "just works, doesn't fall over" state.

frasertweedale:
On Wed, Jan 17, 2018 at 12:30:53AM +0000, Roman Joost wrote:
> Yeah @frasertweedale I agree at this point we don't really need it. Am I in the moot to change it back? Not really. I expect changes to it if we want to actually file a package review to get the package into Fedora. So for now I leave it in a "just works, doesn't fall over" state.
> 
Fair enough.

Issue #151, opened 2018-01-15
ResetTags operation followed by AddTags leaves no tags on mails


romanofski:
Happens on master - 3695c4b2c0dc17b2c2d507744434004be39f25cb

Steps to reproduce:

1) Select a thread or a mail
2) Open the tag editor
3) Use: `= +foo` which should reset all tags and add `foo`

Current result:
Thread or mail is left with no tags applied.

Expected result:
Thread or mail should be tagged with `foo`.

PR #154, opened 2018-01-17
Fix ResetTags op leaving message without tags

romanofski:
The tag operations applied to the data type where applied from right to
left, which has left ResetTag as the last operation.

This patch applies the tag operations from left to right, the same way
we parse them.

Fixes https://github.com/purebred-mua/purebred/issues/151

### DIFF
--- a/test/TestMail.hs
+++ b/test/TestMail.hs
@@ -5,9 +5,11 @@ module TestMail where
 
 import qualified Data.ByteString as B
 import qualified Data.Text.Encoding as T
+import Control.Lens (view)
+import Data.Time.Clock (UTCTime(..), secondsToDiffTime)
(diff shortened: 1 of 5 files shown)


test/TestMail.hs
```diff
@@ -3,24 +3,29 @@
 
 module TestMail where
 
+import Types
+import Storage.Notmuch (addTags, removeTags, tagItem)
+
+import Notmuch (mkTag, tagMaxLen)
+
```
frasertweedale: can you please put "local" deps at bottom of import list?  I like things to run downwards from less specific to more specific imports :)

romanofski:
I'm merging this based on your implicit consent after moving the imports to the bottom of the import list :)

PR #155, opened 2018-01-22
User acceptance tests for the configuration

romanofski:


configs/config.hs
```diff
@@ -1,9 +1,30 @@
...
+--
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-
-Example configuration, which uses more mutt-alike keybindings
+Example configuration, currently used for testing which demonstrates various
+ways to overwrite the configuration.
 -}
 import Purebred
+import qualified Data.ByteString.Lazy as B
```
frasertweedale: I'd prefer `L` for qualifying `Data.ByteString.Lazy`.

romanofski: Done :)

PR #161, opened 2018-02-10
Use latest purebred-email parsing code

romanofski:


stack.yaml
```diff
@@ -54,6 +54,7 @@ extra-deps:
 - data-clist-0.1.2.0
 - word-wrap-0.2
 - vty-5.18.1
+- concise-0.1.0.1
```
frasertweedale: this should get picked up in relevant lts builds very soon.  but yeah whatevs :)

frasertweedale:
@romanofski yeah switch to lts-10 is fine.  We still need to test older LTSes, or switch to plain cabal for testing older GHCs.

PR #162, opened 2018-02-10
read mail files as strict bytestring

frasertweedale:
When parsing using attoparsec, if it is assumed that an input will
parse successfully, and most of the input will be parsed, and the
input is many multiples of the default chunk size for lazy
bytestrings, then strict bytestrings give better performance than
lazy bytestrings.  This is because feeding many chunks to an
attoparsec parser will result in several resizings of the
contiguously allocated buffer.  In contrast, if the input is
provided as a single strict bytestring, there initial (and sole)
buffer reuses the same memory. For MIME messages, most of the parsed
data are bytestrings that also point into the same memory.  Buffer
resizes are avoided and there is minimal copying.

Therefore read emails as strict bytestrings.


romanofski:
Nice. Great explanation!

PR #163, opened 2018-02-10
Upgrade to lts-10.5

romanofski:


README.md
```diff
@@ -6,7 +6,7 @@ An MUA built around [*notmuch*](https://notmuchmail.org/).
 
 ## requirements
 
-- GHC >= 8.0
+- GHC >= 8.2
```
frasertweedale: This is not (yet) the requirement.  We make a point of testing with 8.0, and will do
until all OSes/distros we care about have moved to 8.2 or later.

purebred.cabal
```diff
@@ -65,7 +65,7 @@ library
                      , Purebred
   build-depends:       base >= 4.9 && < 5
                      , lens
-                     , brick >= 0.20
+                     , brick >= 0.34
```
frasertweedale: Big jump.  Are we actually using stuff that requires 0.34?  We should only
tighten version bounds when there's actually an issue with versions  outside the bounds.

romanofski: Yeah I want to use the forms introduced by 0.32. Arguably it should be 0.32 then...

frasertweedale: OK, no worries.  Make a note of that in the commit msg.

README.md
frasertweedale: Can we leave this at 8.0 @romanofski ?  8.2 is not required yet.

frasertweedale: No.  First, you can override the stack resolver that gets used.  Second, you assume developers are using stack at all (on one my systems, where I actually do purebred devel, I am not, and I am using ghc-8.0).  Third, the README is not really talking about requirements for devel, but requirements for using purebred (i.e. build/run).

romanofski: Alright. Fair point. I'll change it back.

frasertweedale: Thanks!

PR #165, opened 2018-02-11
Refactor: Use an empty mail in the compose editor

romanofski:
Instead of starting out with a temporary file, start with an empty mail
which we "fill" with the necessary data to send it.


frasertweedale:
:ship:

PR #166, opened 2018-02-12
implement charset decoding

frasertweedale:
Update to commit of purebred-email that introduces charset decoding
and modify purebred to perform charset decoding when displaying a
message.

Drive-by generalisation of getHeader, getFrom, getSubject and getTo
which work for all instantiations of 'Message s a'.


romanofski:
One test fails, because it matches `Subject\s` (without the colon). Should be a quick fix.

romanofski:
yeah works like a charm 👍

frasertweedale:
\o/

PR #168, opened 2018-02-17
support parameter continuations, encoding and charsets

frasertweedale:
Update to a commit of purebred-email that implements RFC 2231
parameter continuations, encoding and charsets.  Refactor purebred's
email formatting accordingly.


frasertweedale:
Thanks @romanofski.  Next *purebred-email* effort will be for finding/processing attachments.

PR #169, opened 2018-02-17
Fix thread read state not updated

romanofski:


src/Storage/Notmuch.hs
```diff
@@ -218,6 +218,23 @@ threadToThread m = do
...
       <*> Notmuch.threadId m
 
+reloadTags
+    :: (MonadError Error m, MonadIO m, HasId a, ManageTags a)
+    => FilePath -> a -> m a
+reloadTags fp item = withDatabaseReadOnly fp go
+  where
+    go db = getThread db (getId item) >>= Notmuch.tags >>= pure . flip setTags item
+
+class HasId a where
```
frasertweedale: I think this class is a bad idea, because `ThreadId` and `MessageId` - although
they are type synonyms now for `ByteString`, could eventually become different
types via `newtype`.

If you want to keep the class so there is a common access method `getId`, you can
use an associated type:

```haskell
class HasId a where
  type IdType a
  getId :: a -> IdType a

instance HasId NotmuchMail where
  type IdType NotmuchMail = MessageId
  getId = view mailId

...
```

This requires the `TypeFamilies` language extension.

I'm inclined to suggest `getId :: Getter a (IdType a)` too so it can be composed
with other optics.  But I am not fussy about that.

src/Storage/Notmuch.hs
```diff
@@ -218,6 +218,23 @@ threadToThread m = do
       <*> Notmuch.threadTotalMessages m
       <*> Notmuch.threadId m
 
+reloadTags
+    :: (MonadError Error m, MonadIO m, HasId a, ManageTags a)
+    => FilePath -> a -> m a
+reloadTags fp item = withDatabaseReadOnly fp go
+  where
+    go db = getThread db (getId item) >>= Notmuch.tags >>= pure . flip setTags item
```
frasertweedale: [nit] the `pure` is redundant.  you could do:
```haskell
go db = getThread db (getId item) >>= fmap (`setTags` item) . Notmuch.tags
```
or, for slightly better "ordering" i.e. operations all reading right-to-left instead of a mix:
```haskell
go db = fmap (`setTags` item) . Notmuch.tags =<< getThread db (getId item)
```

src/UI/Actions.hs
```diff
@@ -595,3 +601,14 @@ getMailsForThread
...
+reloadThreadTags
+  :: MonadIO m
+  => AppState
+  -> (a, NotmuchThread)
+  -> m (AppState -> AppState)
+reloadThreadTags s (_, thread) =
+  let dbpath = view (asConfig. confNotmuch . nmDatabase) s
+      updateList t' = over (asMailIndex . miListOfThreads) (L.listModify $ const t')
+  in runExceptT (Notmuch.reloadTags dbpath thread)
+     >>= either (pure . setError) (pure . updateList)
```
frasertweedale: This looks like it could be simplified to:
```haskell
let
  ...
in
  either setError updateList <$> runExceptT (Notmuch.reloadTags dbpath thread)
```

romanofski:
Many thanks for the comments. I'll study type families a bit to understand the why and how.

src/Storage/Notmuch.hs
romanofski: Hm... `getId` now returns a type of `IdType a`, but `getThread` wants a `ByteString`. I'm now a bit puzzled how I get my Bytestring out of `IdType` again. Would I need to define a data type or type alias to define another function to pull out the Bytestring?

romanofski: Actually, I'm realising that this can not be generalised. The function calls `getThread` which in turn queries the database with the given thread `id`. While technically it won't fail, passing in a mail ID would return no results and hence doesn't really make sense. Perhaps `getThread` would have to be abstracted so that you can pass in a message- or thread id.

src/Storage/Notmuch.hs
romanofski: Cheers. Done.

src/UI/Actions.hs
romanofski: Is the point of using fmap to simplify that I operate in the Monad instance of the Functor? (Not sure if that sentence makes sense).

src/Storage/Notmuch.hs
romanofski: I removed the class now, since anything which could be abstracted away is perhaps something for another patch.

src/UI/Actions.hs
frasertweedale: yeah, using the weaker abstraction makes it clearer what's going on, i.e. that the next
action cannot depend on the result of the first.  It also avoids two redundant applications of `pure`.

PR #175, opened 2018-06-18
Refactor/flexible ui

romanofski:


romanofski:
I had a look at the stackage nightly failures and fiddled with the versions. I couldn't get something together which works with lts-8 and nightly. So either we ditch the one over the other. Maybe ignore nightly failures for now?

frasertweedale:
@romanofski nightly failure might be due to missing / incompatible package versions.  Next step is to at lts-11 to build matrix, then look at nightly.

PR #184, opened 2018-06-30
Update screenshot to the latest and "greatest"

romanofski:
Our UI has changed since the last screen shot was made. We've added a lot of
additional features as well as UI elements.


frasertweedale:
lulz, I haven't even looked at the pic but whatevs, merge it!

Issue #33, opened 2017-08-11
Theme support


romanofski:
With the use of custom configurations implemented in #8 it should be possible to provide theme support by manipulating the attribute map.

romanofski:
Version 0.29 of brick added [theme](https://github.com/jtdaugherty/brick/blob/master/docs/guide.rst#attribute-themes) support.

romanofski:
I noticed this issue today, but I don't think it's an easy fix. The issue is ambiguous and I suppose I personally would want clarification of how we would want to support themes. Brick allows to define themes by using an `ini` file, yet we configure Purebred entirely through Haskell code. So question I would have with this is:

* Would we support bricks theme support via `ini` file? (Personally not a fan, since it feels awkward from a documentation point of view)
* Should we support adjusting colours via our config? (Seems the most "natural" way, since everything is configured this way too)
* Should we support configuring it via plug-ins? (Not sure how this would look like, but perhaps a won't give us much)

I guess until we have a better understanding I remove the easy fix labels.

frasertweedale:
@romanofski, agree we should look for how to do it via our config.  If we can do that, then plugin support is a natural consequence (it is trivial for a plugin to modify the config at application startup).

PR #188, opened 2018-07-06
Add two themes to choose from

romanofski:
This now adds two themes users can select from: solarizedDark (default) and
solarizedLight. This does not provide the ability to overwrite them with custom
attributes. I decided against it for now, since our executable is already quite
big (24MiB).

Related to https://github.com/purebred-mua/purebred/issues/33


frasertweedale:
Fine with me.  Merge ho!

PR #192, opened 2018-07-14
refactor some Vector-related code

frasertweedale:


src/UI/Actions.hs
```diff
@@ -797,7 +798,7 @@ initialCompose =
         (E.editorText ComposeFrom (Just 1) "")
         (E.editorText ComposeTo (Just 1) "")
         (E.editorText ComposeSubject (Just 1) "")
-        (L.list ListOfAttachments Vector.empty 1)
+        (L.list ListOfAttachments mempty 1)
```
romanofski: Aaah of course.

src/UI/Actions.hs
```diff
@@ -665,7 +666,7 @@ findIndexWithOffset i fx = fmap (i+) . Vector.findIndex fx . Vector.drop i
 
 listSetSelectionEnd :: Lens' AppState (L.List Name a) -> AppState -> AppState
 listSetSelectionEnd list s =
-  let index = Vector.length $ view (list . L.listElementsL) s
+  let index = lengthOf (list . L.listElementsL) s
```
romanofski: hm.. TIL

romanofski:
Looks fine to me. Wonder what exactly of the changes is causing the tests to fail...

frasertweedale:
IKR.  I couldn't see why either.  The failure is consistent.  Well, I'll have a crack at it during today's travels :)

src/UI/Actions.hs
frasertweedale: I worked out the problem; I needed a ` . folded` at the end.  `lengthOf` counts the number of
targets of the optic.  In case of a *lens* it is always ``1``.  But actually `lengthOf` is not efficient *O(n)*.
It's better to `. to length` as the last component in the optic, which will use the `Foldable Vector`
instance for *O(1)* complexity.  Updated commit imminent.

src/UI/Actions.hs
```diff
@@ -731,7 +732,7 @@ manageMailTags
     -> (Int, NotmuchMail)
     -> m (AppState -> AppState)
 manageMailTags s tagop m =
-  either setError updateMails <$> applyTagOps tagop (Vector.singleton m) s
+  either setError updateMails <$> applyTagOps tagop [m] s
```
frasertweedale: @romanofski random comment about this: `(Identity m)` would work fine and, unlike `[]` or `Vector`
has no runtime overhead.  But I was not bothered with it :)

PR #193, opened 2018-07-20
Instant validation feedback

romanofski:
Do not merge.

This is just a little play thing I've been fiddling around with. Purebred now gives you instant feedback if the editors input can be parsed. That is for adding labels/tags and the "From" address when starting to compose mails.
Would you mind having a try tho and tell me if this is useful or just a bit too "loud"?


romanofski:
I'll close this request for now, but keep the branch around to play around with it again.

PR #194, opened 2018-08-06
Use a column width percentage for authors

romanofski:
When purebred renders a list of threads, we render every attribute of a thread
in sequential order. That makes it harder to skim through the threads than it
has to be.

This patch renders all authors in a column with a fixed percentage (see
https://github.com/jtdaugherty/brick/issues/182). This provides a better anchor
point for the eyes to quickly skim through mails.


romanofski:
I adjusted one of the statements which was way to complex for what it tried to assert.

PR #196, opened 2018-09-04
purebred email for sending mail

romanofski:


romanofski:
Here we go ...

src/Config/Main.hs
```diff
@@ -36,6 +38,15 @@ import UI.ComposeEditor.Keybindings
...
 import Storage.Notmuch (getDatabasePath)
 
+sendmailPath :: FilePath
+sendmailPath = "/usr/sbin/sendmail"
+
+renderSendMail :: B.ByteString -> IO String
+renderSendMail m =
+        -- -t which extracts recipients from the mail
+        -- TODO: accept a ByteString as stdin
+        readProcess sendmailPath ["-t", "-v"] (C8.unpack m)
```
frasertweedale: `readProcess` can also block, throw exception. OK for now but file issue to address these problems.

src/UI/Actions.hs
```diff
@@ -758,41 +764,49 @@ replyToMail s =
...
+        (Just m) -> let m' = m
+                          & set (headers . at "Subject") (Just $ T.encodeUtf8 subject)
+                          . set (headers . at "From") (Just $ renderMailboxes from)
+                          . set (headers . at "To") (Just $ renderAddresses to')
+                          . set (headers . at "Date") (Just $ renderRFC5422Date dateTimeNow)
+                    in liftIO $ trySendAndCatch l' (renderMessage $ sanitizeMail m') s
+
+-- Handler to send the mail, but catch and show an error if it happened. This is
+-- really a TODO to improve this, since ideally we can do this in the
+-- cvSendMailCmd. However what prevented me from simply moving there is using
```
frasertweedale: I agree.  We can refactor later.

src/Purebred.hs
```diff
@@ -107,6 +107,21 @@ processConfig =
...
+randomElement xs g = let (i, g') = randomR (0, length xs - 1) g
+                     in (xs !! i, g')
+
+-- | Note: iterate returns a list including the first random character. That
+-- first random character is most often not Boundary compliant, so invoke
+-- iterate with the seed value which returns only boundary compliant characters.
+genBoundary :: RandomGen g => g -> String
+genBoundary g = let seed = randomChar (random g)
+                in fst <$> iterate randomChar seed
+
```
frasertweedale: This whole implementation seems really janky to me.  I'd propose one of two alternatives.

First alternative: define a newtype and instance `Random` for it, then use `randoms`.

Second alternative (simpler):
```haskell
genBoundary :: RandomGen g => g -> String
genBoundary = filter isBoundaryChar . randomRs (minimum boundaryChars, maximum boundaryChars)
  where
  boundaryChars ['0'..'9'] <> ['a'..'z'] <> ['A'..'Z']
  isBoundaryChar = (`elem` boundaryChars)
```

We use `randomRs` as an efficient hack so we don't generate chars outside the range that contains all valid boundary chars.

src/Purebred.hs
```diff
@@ -97,7 +96,8 @@ buildLaunch = do
 
 launch :: UserConfiguration -> IO ()
 launch cfg = do
-    cfg' <- processConfig cfg
+    b <- getStdGen >>= pure . genBoundary
```
frasertweedale: Don't need `(>>=)` here.
```haskell
b <- fmap genBoundary getStdGen
```

src/Config/Main.hs
romanofski: Yeah it seems you saw the comment on the calling site.

src/Purebred.hs
romanofski: I thought the range has to be something like `['a'..'z']`. I also thought why not use:
```Haskell
filter isBoundaryChar . randoms
```
`randoms` would produce the characters and we simply filter what is useful. But boy was that slow. Suppose it needs to be in the given range ...

romanofski: Ouch 😱

configs/config.hs
```diff
@@ -42,13 +42,15 @@ myMailKeybindings =
     [ Keybinding (EvKey (KChar 'a') []) (setTags [RemoveTag "inbox", AddTag "archive"] `chain` continue)
     ]
 
-writeMailtoFile :: Mail -> IO ()
+-- Note: The empty string return value is currently used as an error condition,
+-- but for this case it is not used.
```
frasertweedale: File a ticket to refactor this to return `IO (Either e ())` where `e ~ String` or maybe something else...

src/Config/Main.hs
```diff
@@ -36,6 +38,15 @@ import UI.ComposeEditor.Keybindings
 import Types
 import Storage.Notmuch (getDatabasePath)
 
+sendmailPath :: FilePath
+sendmailPath = "/usr/sbin/sendmail"
+
+renderSendMail :: B.ByteString -> IO String
+renderSendMail m =
+        -- -t which extracts recipients from the mail
+        -- TODO: accept a ByteString as stdin
```
frasertweedale: Yeah, file a ticket for this TODO as well.  Although maybe it can go in the same ticket as the other `readProcess` problems we already discussed.

src/UI/Actions.hs
```diff
@@ -758,41 +764,49 @@ replyToMail s =
...
+-- the MonadError typeclass. Using that, the typevariable has to be carried into
+-- the configuration as another? type variable. So all in all it was non
+-- trivial.
+trySendAndCatch :: String -> B.ByteString -> AppState -> IO AppState
+trySendAndCatch l' m s = do
+    let cmd = view (asConfig . confComposeView . cvSendMailCmd) s
+    catch
+        (cmd m >>= \_ -> pure $ s
```
frasertweedale: Don't need `(>>=)` here.
```haskell
cmd m $> s'
```

configs/config.hs
romanofski: See #200

src/Config/Main.hs
romanofski: See #201. I'll update the comments and repush.

src/Purebred.hs
```diff
@@ -107,6 +110,19 @@ processConfig =
   >=> confEditor id
   >=> (confFileBrowserView . fbHomePath) id
 
+
+-- RFC2046 5.1.1
+boundaryChars :: String
+boundaryChars = ['0'..'9'] <> ['a'..'z'] <> ['A'..'Z'] <> "'()+_,-./:=?"
+
+-- | Note: iterate returns a list including the first random character. That
```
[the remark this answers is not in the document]
romanofski: Ah cheers.

src/Purebred.hs
```diff
@@ -107,6 +110,19 @@ processConfig =
   >=> confEditor id
   >=> (confFileBrowserView . fbHomePath) id
 
+
+-- RFC2046 5.1.1
+boundaryChars :: String
+boundaryChars = ['0'..'9'] <> ['a'..'z'] <> ['A'..'Z'] <> "'()+_,-./:=?"
```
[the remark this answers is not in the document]
romanofski: Done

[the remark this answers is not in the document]
frasertweedale: Got something to push?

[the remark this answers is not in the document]
romanofski: Yes, but let me fix up the parser library first, since I'll have to update this request again with the new SHA

src/UI/Actions.hs
romanofski: Done

src/Purebred.hs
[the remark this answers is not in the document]
romanofski: Done.

src/Purebred.hs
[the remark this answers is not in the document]
romanofski: I tried to make this a `ByteString` but clash with the function which is using it. If you wouldn't mind, I'd defer this and we fix it up later?

romanofski:
Rebased again. Let me know if I have missed anything which prevents merging ...

src/Purebred.hs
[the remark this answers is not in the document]
frasertweedale: Sure, there are other ways to optimise it too.  It's not really important (until we measure it at least).

PR #198, opened 2018-10-07
tests: Bump the notmuch versions

romanofski:
Bump all notmuch versions from 0.24 to 0.26 - 0.27 to test against. Most
distributions ship newer versions and at this point in development there is in
point supporting older notmuch versions.


frasertweedale:
Merge ho!

PR #202, opened 2018-10-12
Use new function to match content type

romanofski:
The ctEq function was deprecated and we want to use the new function to get rid
of the deprecation warning.


frasertweedale:
Merge ho!

PR #203, opened 2018-10-12
Source for the current logo

romanofski:
Not that it's *the* logo, but rather better than the randomly generated
logo from github.


romanofski:
Didn't realise that I never actually pushed this.

frasertweedale:
Merge ho!

Issue #48, opened 2017-08-30
sanity check the configuration


frasertweedale:
We should sanity-check the configuration before launching into the program.

Happy to leave this for someone else for now (maybe do some recruiting at BFPG?).
For now let's just make a list of the things to check.

- [ ] database can be opened
- [ ] initial notmuch search term does not return an insane amount of mail
- [ ] keybindings are non-empty ?

romanofski:
Do you think this is a necessary item for a first release? I'm on the fence about it, since it feels like it is not crucial for the application to work, yet could mean a lot of frustration which could be avoided.

frasertweedale:
On Tue, Sep 12, 2017 at 12:36:34AM +0000, Roman Joost wrote:
> Do you think this is a necessary item for a first release? I'm on the fence about it, since it feels like it is not crucial for the application to work, yet could mean a lot of frustration which could be avoided.
> 
It can wait.  It's a good item for newcomer to project to tackle, if
we get interest.

PR #207, opened 2018-10-14
Feature/configure identities

romanofski:


romanofski:
Suppose we should merge #196 first. This is just an additional commit on top of #196 ...

configs/config.hs
```diff
@@ -52,6 +52,12 @@ writeMailtoFile m = do
   B.writeFile fname m
   pure ""
 
+fromMail :: [Mailbox]
+fromMail =
+    [ Mailbox
+          (Just "Joe Bloggs")
+          (AddrSpec "joe" (DomainDotAtom $ pure "foo.test"))
```
frasertweedale: I'd prefer you created the list with multiple domain parts, i.e.
```haskell
"foo" :| ["test"]
```

If you don't want to import the `NonEmpty((:|))` constructor, you can use `pure` multiple times:
```haskell
pure "foo" <> pure "test"  -- I'm cool with this.
```

src/UI/Actions.hs
```diff
@@ -807,12 +825,12 @@ trySendAndCatch l' m s = do
...
 
-initialCompose :: Compose
-initialCompose =
+initialCompose :: [Mailbox] -> Compose
+initialCompose mailboxes =
   let mail = B.empty
   in Compose
         mail
-        (E.editorText ComposeFrom (Just 1) "")
+        (E.editorText ComposeFrom (Just 1) (decodeLenient $ renderMailboxes mailboxes))
```
frasertweedale: `renderMailboxes` will probably render non-ASCII content using encoded-word, which is probably not what you want.
More likely you want to render the `Mailbox`es directly as `Text`?  (You would need to write that render function).

configs/config.hs
frasertweedale: `NonEmpty` is in *base*.  But as I said I don't mind either way.

src/UI/Actions.hs
romanofski: That reminds me: https://github.com/purebred-mua/purebred-email/pull/15#discussion_r224306043 I didn't implement it. Perhaps I'll do that first and think of a pretty printing function.

romanofski: Hm... you mentioned in purebred-mua/purebred-email#15 that you're working on encoded-word support for address headers. I suppose I'll wait until this is merged before I can continue here?

frasertweedale: @romanofski I am working on that, but I don't think you're blocked.  You should focus on pretty printing and parsing the `Address` (and related types, if necessary) to/from `Text`.  The encoded-word work focuses on the 7bit-compatible conversion to/from ByteString for use directly within RFC822/RFC5322 messages.

romanofski:
Ignore this push please. It's a rebase and addressed the config. Does not include the changes re from field.

src/UI/Actions.hs
romanofski: Hm... there is something missing I don't understand.

At this point, I go from `Mailbox` to `ByteString` to `Text`. Your proposal is to pretty print it from `Mailbox` to `Text`. That seems to be the easy part.
What I'm missing is the parser. Where would a parser which parses a `Text` instead of a `ByteString` be different? If I currently try to stick a:

```Haskell
"\"Róman Joost\" <roman@host>" :: ByteString
```
into the parser I get:
```
Left "\": Failed reading: satisfy"
```
Understood, because the display name contains non-ascii characters which will have to be encoded. In order to make the parser work with Text, I'd basically replace the necessary small bits to parse a `Text` instead of a `ByteString`. That sounds like duplication? Where is my mistake?

frasertweedale: There is no mistake; we'll have to duplicate some stuff.

romanofski: I've went down the rabbit hole and implemented most of the parser actions, which were rarely changes necessary apart from the different type functions. Perhaps we'll have to expose this module with some reshuffling? Something like:

```Haskell
-- the current parser
Data.RFC5322.ByteString.Address
-- and
Data.RFC5322.Text.Address
```

frasertweedale:
> I pushed my changes to: https://github.com/purebred-mua/purebred-email/tree/feature/address_parsing_text which includes a parser for addresses using the `Text` type. Now I realise, that it's probably not really what we want. It wouldn't parse non-ascii addresses like `"Róman Joost" [roman@host](mailto:roman@host)" because quoted text is still ascii only. I guess it depends, since we'll need to able to encode it if it's non-ascii.

It is what we want, because right now it is for dealing with user input.  If purebred-email is not encoding addresses with non-ASCII chars properly, that's a separate issue to deal with.

> 
> Also, I've stumbled over: https://tools.ietf.org/html/rfc6531#section-3.3 which could be useful?
>
RFC 6531 is not used much in practice.  YAGNI (for now).

> Anyhow, depending how we go, at this point:
> a) Is this parser what we want and we deal with the encoding of non-ascii for both the `ByteString` and `Text` variants later?

Yes.

> b) do we need to use a better module path other than `RFC5322.Address.Text`?

Seems fine to me.

> c) perhaps we could use the `mailboxFixtures` which currently use `ByteString` for testing both the `ByteString` and `Text` parsers? I'd simply keep the `ByteString` variant, but for the `Text` bit, use a `decodeLenient`. Perhaps use a separate Test module for both parsers and move the fixtures used by both in their own module?

Should be able to generalise the type of `mailboxFixtures` to provide both cases.
Then we only need to add a few specific tests for each parser, namely, addresses
with encoded-words for the `ByteString` parser, and addresses with unicode
characters for the `Text` parser.  Hopefully the errors are the same for each parser but
if it, I don't mind if we do not assert the exact error message::

```haskell
commonMailboxFixtures
  :: (IsString s)
  => [(String, Either String Mailbox -> Assertion, s)]
commonMailboxFixtures
  -- the current body of mailboxFixtures, except just use
  -- string literals for the input field.  OverloadedStrings will
  -- take care of conversion to ByteString or Text
```

frasertweedale:
@romanofski further to previous post, I see from the commit that the error messages are all (or almost all) the same, so the technique should work.

romanofski:
Hm.. there is one tiny detail with the types tho. Currently we have:
```Haskell
-- Data.RFC5322
data Mailbox =
    Mailbox (Maybe T.Text {- display name -})
             AddrSpec
    deriving (Show,Eq)

data AddrSpec =
    AddrSpec B.ByteString {- local part -}
             Domain
    deriving (Show,Eq)

data Domain
    = DomainDotAtom (NonEmpty B.ByteString {- printable ascii -})
    | DomainLiteral B.ByteString
    deriving (Show,Eq)
```
but now we also have:

```Haskell
-- Data.RFC5322.Address.Text
data Mailbox =
    Mailbox (Maybe T.Text {- display name -})
             AddrSpec
    deriving (Show,Eq)

data AddrSpec =
    AddrSpec T.Text {- local part -}
             Domain
    deriving (Show,Eq)

data Domain
    = DomainDotAtom (NonEmpty T.Text {- printable ascii -})
    | DomainLiteral T.Text
    deriving (Show,Eq)
```
Would it make sense to use a type variable for `Domain` and `AddrSpec`? Otherwise we'll have to convert back and forth all the time. Perhaps it's not that much of a problem. I just stumbled over it when adjusting the tests, since one is the input, but the assertion uses `Data.RFC5322.Mailbox` which is currently only a `ByteString`.

frasertweedale:
@romanofski the same data is represented in both cases.  There should be a single definition of those data types (the definitions in `Data.RFC5322` seem correct to me), with the two sets of parsers/printers to/from `Text` and `ByteString` working against the shared data type definitions.

configs/config.hs
```diff
@@ -52,6 +53,12 @@ writeMailtoFile m = do
   B.writeFile fname m
   pure ""
 
+fromMail :: [Mailbox]
+fromMail =
+    [ Mailbox
+          (Just "Joe Bloggs")
+          (AddrSpec "joe" (DomainDotAtom $ "foo" :| ["test"]))
+    ]
```
frasertweedale: It could be useful to write an `IsString` instance for Mailbox, that just blows up if it can't be converted.
This means a runtime error when processing the configuration (so we should `deepseq` / `force` the whole configuration value to catch it at program startup) but it would be more user-friendly IMO.  LMK what you think.
(Work for a subsequent ticket IMO; and easyfix, so something a new contributor could tackle).

romanofski: Yeah I like the idea. You did actually file a ticket for configuration validation (#48), but I suppose this would be a smaller item than that. Would you mind perhaps to file a ticket?

frasertweedale: - https://github.com/purebred-mua/purebred-email/issues/18
- https://github.com/purebred-mua/purebred/issues/212

PR #210, opened 2018-11-16
avoid runs of multiple (<+>) and (<=>)

romanofski:
The (<+>) and (<=>) combinators create nested structures that are
less efficient than a flat hBox/vBox containing the widgets.
Convert such occurrences to hBox/vBox.

Also remove the VBox newtype whose Semigroup/Monoid instances would
be inefficient.

Also do a bit of drive-by refactoring and add commentary about other
possible optimisations to try in future.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -8,11 +8,9 @@ module Types
   , Tag
   ) where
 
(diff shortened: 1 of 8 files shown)


frasertweedale:
Merge when green.

Issue #124, opened 2017-12-12
Simple version which build/version of purebred users are running


romanofski:
I think purebred should have an option to see which version is running, in order to associate bug reports with a version.

Something like:
```
$ purebred --version
0.1.0.0 unreleased whatever
```
Hint: check for:
```Haskell
import Data.Version (showVersion)                                                              
import Paths_purebred (version) -- generated by cabal
```

frasertweedale:
Good idea. Can show it in *help* view too, when the app is running.

romanofski:
Actually for that won't we need an 'about' screen?

frasertweedale:
/me shruggs.  Doesn't really matter.  Chuck it in the status bar for all I care right now :)

romanofski:
Alright, whoever picks it can choose.

romanofski:
Should perhaps include the hs-notmuch version introduced in purebred-mua/hs-notmuch#26

frasertweedale:
@romanofski do we consider this done now?

romanofski:
Yeah... suppose if we want to show the version in the UI we can add this a separate item.

PR #215, opened 2018-11-30
Show version information

romanofski:
With this patch, users can invoke the program with --help to get version
information.

Part of: https://github.com/purebred-mua/purebred/issues/124

### DIFF
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -114,6 +114,8 @@ import Data.Semigroup ((<>))
 import System.Environment (lookupEnv)
 import System.FilePath.Posix ((</>))
 import System.Random (RandomGen, getStdGen, randomRs)
+import Data.Version (showVersion)
+import Paths_purebred (version)
 
 import UI.Index.Keybindings
 import UI.Mail.Keybindings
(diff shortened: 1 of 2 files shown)


romanofski:
I want to create a new Fedora build, but would like to add in the version information. This patch doesn't show version information in the UI, but only if you invoke --help. I first thought about also implementing a way to show it in the UI, but ignored it for now.

src/Purebred.hs
```diff
@@ -145,7 +147,11 @@ appconfig =
 optParser :: ParserInfo AppConfig
 optParser = info
   (appconfig <**> helper)
-  (fullDesc <> progDesc "purebred" <> header "a search based, terminal mail user agent")
+  (fullDesc
+   <> progDesc "purebred"
+   <> header ("a search based, terminal mail user agent - " <> versionString))
+  where
+    versionString = showVersion version
```
frasertweedale: We might as well add a `--version` option while we're at it?

romanofski: Fair enough... will update

romanofski:
Alright here we go. Was tempted to include more information from https://hackage.haskell.org/package/Cabal-2.2.0.1 but then shied away from depending on it.

frasertweedale:
@romanofski what other information is available with Cabal-2.2?

Anyhow, ACK.  There are some tweaks I want to make bug will tackle later.

romanofski:
Ooh.. `buildArch` was one I thought would be interesting ...

PR #216, opened 2018-12-01
Update spec file for new Fedora/Copr build

romanofski:
Update all changed dependency versions and dependencies since the last
successful build. Also adds the URL to the sources which currently point to
github instead of hackage.


romanofski:
Successful build: https://copr.fedorainfracloud.org/coprs/romanofski/purebred/build/832033/ Which means Fedora 28, 29 and rawhide now delivery latest snapshot

frasertweedale:
ACK; merge when green.

PR #219, opened 2018-12-08
add lazy vector data type

frasertweedale:
Add a data type that can be used to load search result lazily.  The
changes to actually use it will be made in subsequent commits.

Bump brick lower bound to 0.44, which is the first version with
Brick.Widgets.List.GenericList.


frasertweedale:
nak, need to tweak cabal file

romanofski:
I've restarted the failed test run for now, since I think I've found the problem.

PR #220, opened 2018-12-10
improve recompile with cabal new-style install

frasertweedale:
When running purebred cabal new-style install and recompile is
needed, GHC needs a bit of help to find the purebred package.  When
constructing ghcOpts, inspect the libdir to see if it looks like
we're running from a cabal new-style install, and if so extract the
package-id from it.


romanofski:
Feels a bit hacky, but if you think it's reliable and works, merge when green :)

Issue #204, opened 2018-10-13
Test timeout before purebred started


romanofski:
One test failed today with the following output:
```
use file browser to add attachments:                      FAIL (5.05s)
      Wait time exceeded. Condition not met: 'Literal "Purebred: Item"' last screen shot:
      
       export TERM=ansi
      export GHC=stack
      export GHC_ARGS="$STACK_ARGS ghc --"
      export PUREBRED_CONFIG_DIR=/tmp/purebredtest8989
      purebred --database /tmp/purebredtest8989/Maildir/
      travis@travis-job-fddae86d-6b88-4787-a62a-052ac37ea9b4:~/build/purebred-mua/pure
      bred$ export TERM=ansi
      travis@travis-job-fddae86d-6b88-4787-a62a-052ac37ea9b4:~/build/purebred-mua/pur
      ebred$ export GHC=stack
      travis@travis-job-fddae86d-6b88-4787-a62a-052ac37ea9b4:~/build/purebred-mua/pur
      ebred$ export GHC_ARGS="$STACK_ARGS ghc --"
      travis@travis-job-fddae86d-6b88-4787-a62a-052ac37ea9b4:~/build/purebred-mua/pur
      ebred$ export PUREBRED_CONFIG_DIR=/tmp/purebredtest8989
      travis@travis-job-fddae86d-6b88-4787-a62a-052ac37ea9b4:~/build/purebred-mua/pur
      ebred$ purebred --database /tmp/purebredtest8989/Maildir/
      Selected resolver: lts-9.21
      
      
      
      
      
      
      
      
      
[...]

PR #221, opened 2018-12-10
refactor WithContext instances for list views

frasertweedale:
The 'WithContext' instances for the various brick 'List' states can
be replaced by a single, more general instance.  We can also remove
the (Eq e) constraint from 'currentItemW'.

### DIFF
--- a/src/UI/Status/Main.hs
+++ b/src/UI/Status/Main.hs
@@ -45,16 +45,7 @@ statusbar s =
 class WithContext a where
   renderContext :: AppState -> a -> Widget Name
 
-instance WithContext (L.List Name NotmuchThread) where
-  renderContext _ = currentItemW
-
-instance WithContext (L.List Name NotmuchMail) where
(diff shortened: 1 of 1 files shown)


romanofski:
Good stuff!

frasertweedale:
@romanofski just laying some ground work so that my eventual PR to use the lazy list stuff is simpler.

romanofski:
Hmmm spurious failures? Haven't had a thorough check...

romanofski:
Looks like you ran into the same problem I documented in #204. So it existed with the code not using dire. I first thought the compiling of the config would take a bit longer than 5s, but I almost doubt that. The stack tool determining it will have to recompile something?

frasertweedale:
yeah, the compile could totally take longer than 5s.  Maybe the first test we do should just be to run the program and wait for it to paint the UI, with a long time (even up to 60s).  After that we should be good.

PR #222, opened 2018-12-10
remove unneeded export

frasertweedale:


romanofski:
Jeez... it really takes a long time for travis to pick up the merge request; test run itself is short in comparison. Wondering if there is something we can tweak or just have to accept the poor scheduling time ...

PR #223, opened 2018-12-11
tests: Exponentially back off waiting

romanofski:
This patch adds an exponential back off to the wait times when waiting
for a repaint of the UI. With the default count down of 5 we wait:

	Prelude> [(20*10^3)*x | x<-[4,8,12,16,20]]
	[80000,160000,240000,320000,400000]

up overal up to 1.2s for the UI to be repainted for every step. This is
shorter to the previous setting which waited for every step 0.5s which
was cumulatively 2.5s until it bailed out.

This has the potential to shorten the test run time from 98s down to
30s.

Note: The UI repaint depends on the IO rate of the machine the tests
run. No consideration has taken into account to determine the IO rate
for an optimal initial back off time with this patch.

### DIFF
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -765,7 +765,7 @@ sendKeys :: String -> Condition -> ReaderT Env IO String
 sendKeys keys expect = do
     sessionName <- getSessionName
     liftIO $ callProcess "tmux" $ communicateSessionArgs sessionName keys False
-    waitForCondition expect defaultCountdown
+    waitForCondition expect defaultCountdown initialBackoffMicroseconds
 
 sendLiteralKeys :: String -> ReaderT Env IO String
 sendLiteralKeys keys = do
@@ -784,8 +784,8 @@ getSessionName = view (envSessionName . ask)
 getTestMaildir :: (Monad m) => ReaderT Env m FilePath
 getTestMaildir = view (envMaildir . ask)
 
-holdOffTime :: Int
-holdOffTime = 10 ^ (6 :: Int)
+initialBackoffMicroseconds :: Int
+initialBackoffMicroseconds = 20 * 10 ^ (3 :: Int)
 
 -- | convenience function to print captured output to STDERR
 debugOutput :: String -> IO ()
@@ -794,11 +794,15 @@ debugOutput out = do
   when (isJust d) $ hPutStr stderr ("\n\n" <> out)
 
 -- | wait for the application to render a new interface which we determine with
---   a given condition. We check up to @n@ times, waiting a short duration
---   between each check, and failing if the tries exhaust with the condition
---   not met.
-waitForCondition :: Condition -> Int -> ReaderT Env IO String
-waitForCondition cond n = do
+--   a given condition. We wait a short duration and increase the wait time
+--   exponentially until the count down reaches 0. We fail if until then the
+--   condition is not met.
+waitForCondition ::
+ Condition
+ -> Int  -- ^ count down value
+ -> Int  -- ^ milliseconds to back off
+ -> ReaderT Env IO String
+waitForCondition cond n backOff = do
   out <- capture >>= checkPane
   liftIO $ assertBool
     ( "Wait time exceeded. Condition not met: '" <> show cond
@@ -811,8 +815,8 @@ waitForCondition cond n = do
       | checkCondition cond out = pure out
       | n <= 0 = pure out
       | otherwise = do
-          liftIO $ threadDelay holdOffTime
-          waitForCondition cond (n - 1)
+          liftIO $ threadDelay backOff
+          waitForCondition cond (n - 1) (backOff * 4)
 
 checkCondition :: Condition -> String -> Bool
 checkCondition (Literal s) = (s `isInfixOf`)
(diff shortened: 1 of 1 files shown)


romanofski:
Test run time from 116s down to 56s, which is not too bad. Although I'd be more interested if it actually addresses the spurious failure ...

frasertweedale:
AFAICT, on each recursion the duration gets mutliplied by 4, i.e. it is not a linear increase as
you suggest in the commit message but rather:

```
>>> 20000 / 1e6
0.02
>>> _ * 4
0.08
>>> _ * 4
0.32
>>> _ * 4
1.28
>>> _ * 4
5.12
```

I'm fine with the implementation but the commit message should match it :)

I don't think this will help with the spurious failure due to recompile.  For that, I think we need to make the first test have a really long timeout to allow for recompilation time.

frasertweedale:
@romanofski ACK when commit message updated to match code.

romanofski:
I updated this patch now with the addition to re-compile the configuration during setup. Just realizing however, that the invocation of the purebred command will probably fail. Since we never install the compiled binary. Will update ...

romanofski:
Ah dang.. we do install the binary:
```
stack $STACK_ARGS --no-terminal install
```
but not the configuration. That'll probably not do what I thought it'll do :/ Will have to extend the function...

frasertweedale:
@romanofski are we good?  I'm happy so if you're happy, merge ho.

romanofski:
No wait. There was a problem with the compilation. I'd like to check that it actually did the right thing.

romanofski:
Hm.. not convinced:
```
$ PUREBRED_CONFIG_DIR=./configs stack exec purebred -- --version
Downloaded lts-10.5 build plan.
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Downloaded ghc-8.2.2.
Installed GHC.
Configuration './configs/purebred.hs' changed. Recompiling.
Error occurred while loading configuration file.
0.1.0.0
The command "PUREBRED_CONFIG_DIR=./configs stack exec purebred -- --version" exited with 0.
```
It looks like it's doing the right thing, but I doubt that it has created the cached binary. Although now thinking of it, I could test for it. Anyway, I'll have to check tomorrow again with the hope to reproduce the error.

frasertweedale:
@romanofski upon further investigation, I see that this approach won't work.
The acceptance tests put the config file in a temporary directory and supply
`PUREBRED_CONFIG_DIR` envvar.  Each test cases uses a new temp dir with
the config file copied into it (every `withTmuxSession`).  The binary will (or should)
be recompiled each test case (because even though each config is the same, purebred
doesn't know that; they might be different).

Furthermore, regardless of what we do on Travis CI, developers running the test suite on a
slow machine might run into timeout issues.  I think the solution needs to be part of the test
suite itself.  A couple of ideas:

1. make executing `purebred --version` part of the `setUp` routine (still recompiles for each test case but it's happens before the actual tests, so timeout is not a concern)
2. use a single config dir but a different `--database` option for the different test cases (recompile once for all tests; can be done in the `pre` action in `systemTests`).

As for why that travis step is failing, I'm not sure, and the output doesn't give much of a clue.
But per preceding comments I think we need to change track anyway.

Sidenotes: I want to get rid of stack stuff and just use cabal in the test suite.  Not a high priority but I think things will go smoother.  Also we need to cache the hell out of our artifacts, CI is taking waaaay too long.

romanofski:
heh - my initial thoughts exactly:
```
14:16 < romanofskiWork> ftweedal: hey just wondering why you'd want a single test to cater for a longer recompilation time? Since the config (although the same) is used for each 
                        test, wouldn't that mean that the config is recompiled new for each test?
14:17 < ftweedal> it should be cached right?
14:18 < ftweedal> i.e. the binary artifact gets cached and recompilation occurs only if the modification type of the "entry point" executable or the config file exceeds the 
                  modification time of the cached binatry
```

I can a) reproduce the error `Error occurred while loading configuration file` locally and b) will look into the option to force a recompile for each test setup. Reason being is, that I think we may want to use separate config files for separate test cases any ways.

I guess until this actually works we better not merge it. There is no way really to prevent that in github isn't it?

romanofski:
Seems like it fails to construct the path to the cache dir if I use a relative path:
```
$ PUREBRED_CONFIG_DIR=configs stack exec purebred -- --version
"user config: configs"
Configuration 'configs/purebred.hs' changed. Recompiling.
Error occurred while loading configuration file.
0.1.0.0
```
in contrast to an absolute path:
```
PUREBRED_CONFIG_DIR=$(pwd)/configs stack exec purebred -- --version
"user config: /home/rjoost/works/purebred/configs"
Configuration '/home/rjoost/works/purebred/configs/purebred.hs' changed. Recompiling.
Program reconfiguration successful.
Launching custom binary /home/rjoost/works/purebred/configs/purebred-linux-x86_64

"user config: /home/rjoost/works/purebred/configs"
0.1.0.0
```
Also confirmed that creating new config directories will not help at all. Reason why it confused me was that I thought that the cache directory might be the same, hence the binary for the same config file may stay the same.

That cleared, I look into moving the compilation step into a test case setup.

frasertweedale:
@romanofski the backoff patch is fine, maybe just remove the other patch from this PR and we can merge it.

romanofski:
@frasertweedale I like the idea of having a separate test suite. Have a crack at it if you want or file a new backlog item. Either way is fine :)

romanofski:
Alright... updated with the exponentialBackoff patch only.

PR #224, opened 2018-12-12
More preparation for the lazy list work.  There will probably be a few more of these PRs before the main event.

frasertweedale:
Changes:

```
09eeea0 (Fraser Tweedale, 23 minutes ago)
   generalise 'seekIndex'

   Update 'seekIndex' to work not only for 'Vector' but any container type
   that is 'Foldable' and 'Splittable'.

50abeee (Fraser Tweedale, 30 minutes ago)
   add generic 'updateAtIndex' function

   'manageMailTags' uses 'safeUpdate' which was specific to Vector. Alter it
   to use a more general version that works for any
   'Traversable' and remove 'safeUpdate'.

147cdd4 (Fraser Tweedale, 37 minutes ago)
   use 'mempty' to construct initial state

   To smooth the way for later changes, remove mentions of 'Vector' from
   UI.App and rely in the Monoid instance instead.
```

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -45,6 +45,8 @@ module UI.Actions (
   , delete
   ) where
 
+import Data.Functor.Identity (Identity(..))
+
 import qualified Brick
 import qualified Brick.Focus as Brick
 import qualified Brick.Types as T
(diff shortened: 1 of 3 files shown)


romanofski:
> before the main event

Haha almost sounds like you're building up to something like a symphony.

PR #225, opened 2018-12-12
reset list when applying search

frasertweedale:
When applying a search we currently try and preserve the currently
selected item.  This doesn't make a lot of sense - it will usually
be a different item.  Furthermore when using a lazy list type it
will force evaluation of the result up to the current index
(undesirable).

Therefore, when applying a search always set the selected index back
to zero.

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -72,7 +72,6 @@ import System.FilePath (takeFileName, takeDirectory, (</>))
 import qualified Data.Vector as Vector
 import Prelude hiding (readFile, unlines)
 import Data.Functor (($>))
-import Control.Applicative ((<|>))
 import Control.Lens
        (_Just, to, at, ix, _1, _2, toListOf, traversed, has, snoc,
         filtered, itoList, set, over, preview, view, (&), nullOf, firstOf,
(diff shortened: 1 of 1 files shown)


romanofski:
I think the point being is to keep the list position instead of always jumping to the top. The reason for keeping the position is because when you archive mails, you typically go from top to bottom to figure out what can be archived. That follows a little bit the form of a todo list. When you've tagged mails to be archived it is good to not show them any more and performing a new search basically hides them. If the cursor jumps back to the initial position you repeatedly have to find where you left off again if your inbox has a bit of mails in it.

I'm sure this could be perhaps accomplished differently (maybe hide the archived items instantly when they're tagged as such). Just wanted to point out the motivation as to why it was the way it was.

So merge when green.

romanofski:
> @romanofski only problem is it keeps you at the same _index_, not the same _item_.
> There is a good chance you end up nowhere near where you were before, right?
That's right. The point however is not that you want to stick to the same item. The point is that you remain in a similar position as to where you left off with the tagging. If your cursor is reset every time to it's initial position, you always need an additional scroll step to go where you were. How do I know where I was? Well I do remember the surrounding mails based on their subjects.

> You can seek the previous item (if it is still present in the list) fairly easily at _O(n)_ cost,
> i.e. seek by thread ID or message ID. Anyhow if you find the new behaviour annoying
> we can open a ticket to discuss whether or how to change the behaviour.
Yep sounds good.

PR #226, opened 2018-12-12
simplify tag management internals

frasertweedale:
The code to deal with updating tags and updating the selected list item is
not great.  Especially the parts dealing with ((,) Int). Use
Brick.Widgets.List.listModify to modify the currently selected item,
avoiding the need to pass the index around.

Also simplify the 'haveTagsChanged' implementation while driving through.

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -5,6 +5,9 @@
 -- | module for integrating notmuch within purebred
 module Storage.Notmuch where
 
+import Control.Monad (when)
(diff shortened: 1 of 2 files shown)


romanofski:
Thank you for this!

frasertweedale:
hm, something is busted...

frasertweedale:
Ah, I refactored `haveTagsChanged` wrong, from ``foo a /= foo b`` to ``(==) `on` foo``.  Whups!

PR #227, opened 2018-12-14
tests: move UAT into separate test suite

frasertweedale:
Add a second test-suite called "uat" and move TestUserAcceptance
into it.  The main advantage is so that you can run just the unit
tests via 'cabal (new-)test unittests' and avoid the expensive UAT
suite.


romanofski:
Hm... One job failed. Would want to check if the timeout is too aggressive...

PR #229, opened 2018-12-14
Revert "tests: Exponentially back off waiting"

romanofski:
This reverts commit cb94323f8d7442e5f59f4169afc38b388d2ca61e.

The more aggressive (read shorter) initial timeouts could lead to more
failing tests. This is partially because after we've implemented the new
UI methods, we never investigated that the views now share more common
widgets. Widget labels show up in more different views. This leads to
more failures in combination with the aggressive timeouts, because the
assertions still find the same strings in the old views as well as in
the new views.

Therefore, tweaking more with the timeouts will make the tests more
brittle.

### DIFF
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -765,7 +765,7 @@ sendKeys :: String -> Condition -> ReaderT Env IO String
 sendKeys keys expect = do
     sessionName <- getSessionName
     liftIO $ callProcess "tmux" $ communicateSessionArgs sessionName keys False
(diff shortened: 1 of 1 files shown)


romanofski:
@frasertweedale Thought it might be better reverting this for now, since it'll make the tests more brittle. I think we'll need a more reliable way to figure out when the UI is repainted. Not sure what that might be.

romanofski:
I take this out for now. No point of speeding things up when they make the tests more unreliable.

PR #231, opened 2018-12-18
Tests/precompile config

romanofski:


test/TestUserAcceptance.hs
```diff
@@ -20,41 +20,48 @@
...
 import Data.MIME (parse, message, mime, MIMEMessage)
 
 {-# ANN module ("HLint: ignore Reduce duplication" :: String) #-}
 
 -- | A condition to check for in the output of the program
 data Condition
-  = Literal String
-  | Regex String
+  = Literal T.Text
+  | Regex T.Text
```
romanofski: Wasn't sure if I should keep it `T.Text` or use `Regex`. I did like the fact, that I can easily just add more text on the `buildAnsiRegex` return value. What do you think?

test/TestUserAcceptance.hs
```diff
@@ -693,7 +709,10 @@ setUpTempMaildir = do
 -- | run notmuch to create the notmuch database
 -- Note: discard stdout which otherwise clobbers the test output
 setUpNotmuch :: FilePath -> IO ()
-setUpNotmuch notmuchcfg = void $ readProcess "notmuch" ["--config=" <> notmuchcfg, "new"] []
+setUpNotmuch notmuchcfg = let config = setStdin closed
+                                       $ setStdout closed
```
frasertweedale: Is there a problem with the default process config (inherited stdin/out/err)?
Put another way, what's wrong with:
```haskell
void $ readProcess_ $ proc "notmuch" args
```
(which is the approach taken at most other points in this diff...)

test/TestUserAcceptance.hs
```diff
@@ -670,14 +676,24 @@ tearDown (Env testdir _ sessionName) = do
...
   setUpPurebredConfig testdir
+  precompileConfig testdir
   pure $ Env testdir maildir sessionName
 
+precompileConfig :: FilePath -> IO ()
+precompileConfig testdir = do
+  env <- getEnvironment
+  let purebredConfig = unionBy (\x y -> fst x == fst y) [("PUREBRED_CONFIG_DIR", testdir)] env
+      config = setEnv purebredConfig $ proc "stack" ["exec", "purebred", "--", "--version"]
+  _ <- readProcessWithErrorOutput_ config
```
frasertweedale: Why do you use `readProcessWithErrorOutput_` when we throw the result away?

purebred.cabal
```diff
@@ -130,20 +130,21 @@ test-suite unittests
...
                      , tasty-hunit
                      , directory
-                     , process
+                     , typed-process
                      , bytestring
                      , ini
                      , temporary
                      , text
-                     , regex-posix
+                     , text-icu
```
frasertweedale: why is it necessary to change from regex-posix to text-icu?

test/TestUserAcceptance.hs
```diff
@@ -898,3 +917,15 @@ buildAnsiRegex attrs fgs bgs =
...
   in
     choice tmux24 tmux25
+
+-- | Interleaves stderr with stdout
+-- Does not accept STDIN arguments, since we do not need to communicate with the
+-- process
+readProcessWithErrorOutput_ :: ProcessConfig stdinClosed stdout stderr -> IO T.Text
+readProcessWithErrorOutput_ pc = do
+  (_, out) <- withProcess_ config $ \p -> atomically $ (,) <$> waitExitCodeSTM p <*> getStdout p
+  pure $ T.decodeUtf8 $ LB.toStrict out
```
frasertweedale: Danger, Will Robinson!

This is non-total.  If there are ANSI escape codes etc, won't this fail (when the return `T.Text` value is demanded)?
Shouldn't we return a `ByteString` and let the caller decide what to do with it?

romanofski:
Cheers. Great questions. Might have to verify with typed-process, but I'll answer the ones I can answer from memory

test/TestUserAcceptance.hs
romanofski: I think by default stderr is not closed and it interleaves with the tasty output. I'll check if it was changed by mistake...

test/TestUserAcceptance.hs
romanofski: Hm ... yeah. Will change.

purebred.cabal
romanofski: Because I had the plan to change from `String` to `Text`. Hm... Maybe it's a bit overkill...

test/TestUserAcceptance.hs
frasertweedale: `readProcess_` creates pipes for the child's stdout and stderr.

purebred.cabal
frasertweedale: Yeah, let's not do that as part of this effort, if it can possibly be avoided.

purebred.cabal
```diff
@@ -130,20 +130,21 @@ test-suite unittests
 test-suite uat
   type:                exitcode-stdio-1.0
   hs-source-dirs:      test
-  ghc-options:         -Wall
+  ghc-options:         -Wall -threaded
```
frasertweedale: :+1:

test/TestUserAcceptance.hs
```diff
@@ -670,14 +676,24 @@ tearDown (Env testdir _ sessionName) = do
...
   prepareEnvironment sessionName
   (testdir, maildir) <- setUpTempMaildir
   setUpPurebredConfig testdir
+  precompileConfig testdir
   pure $ Env testdir maildir sessionName
 
+precompileConfig :: FilePath -> IO ()
+precompileConfig testdir = do
+  env <- getEnvironment
+  let purebredConfig = unionBy (\x y -> fst x == fst y) [("PUREBRED_CONFIG_DIR", testdir)] env
```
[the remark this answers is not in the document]
romanofski: Ah yeah... heh *doh :) Cheers.

romanofski:
Alright. I hope I caught every bit :)

test/TestUserAcceptance.hs
```diff
@@ -835,7 +850,7 @@ startApplication = do
   sessionName <- getSessionName
   -- add $STACK_ARGS so that we use the same resolver as was used for the build
   let cmdline = "stack $STACK_ARGS exec purebred -- --database " <> testmdir <> "\r"
-  liftIO $ callProcess "tmux" $
+  liftIO $ runProcess_ $ proc "tmux" $
```
frasertweedale: (nit) I don't think we need the `liftIO` anymore (also in a bunch of other places).  Don't bother fixing now, unless there's something else to do.

romanofski: Ah good point actually. I thought I saw it somewhere else too. I'll try to keep it in mind for the next patch set, since I want to try out if I can make the tests more reliable when it comes to figuring out if the UI has been repainted.

Issue #201, opened 2018-10-09
Replace System.Process with System.Process.Typed


romanofski:
We should evaluate replacing the current use of `System.Process` with `System.Process.Typed` from http://hackage.haskell.org/package/typed-process-0.2.3.0/docs/System-Process-Typed.html

See discussion on #196

frasertweedale:
This was done for UAT in https://github.com/purebred-mua/purebred/commit/078db1de275fa0026738f579af3b54da02606f4d.

frasertweedale:
Nooope, we still got a bunch of other places to update it :)

romanofski:
Ah sorry.. yes! of course.. the main program...

Issue #200, opened 2018-10-09
mail send command should return a proper error type


romanofski:
With fixing #196 we introduced a new command for sending mail with the type signature:

```Haskell
B.ByteString -> IO String
```
It is used to take the serialised mail and invoke the `sendmail` command for it. The `String` value is a poor man's way of indicating an error.

This should be refactored and a proper error type used:
```Haskell
IO (Either e ())
```
The `e` can be a `String` here, but possibly better is to define a new `Error` constructor.

frasertweedale:
Ultimately, sending mail should have a plugin interface so folks can send mail via SMTP or whatever.

PR #235, opened 2018-12-20
Replace process with typed-process

romanofski:
This removes the dependency from the process package in favour of the
typed-process package. It provides more expressive types to interact with the
system.

Fixes https://github.com/purebred-mua/purebred/issues/201

### DIFF
--- a/configs/purebred.hs
+++ b/configs/purebred.hs
@@ -43,15 +43,13 @@ myMailKeybindings =
     [ Keybinding (EvKey (KChar 'a') []) (setTags [RemoveTag "inbox", AddTag "archive"] `chain` continue)
     ]
 
--- Note: The empty string return value is currently used as an error condition,
--- but for this case it is not used.
-writeMailtoFile :: B.ByteString -> IO String
+writeMailtoFile :: B.ByteString -> IO (Either Error ())
 writeMailtoFile m = do
   confdir <- lookupEnv "PUREBRED_CONFIG_DIR"
   currentdir <- getCurrentDirectory
   let fname = fromMaybe currentdir confdir </> "sentMail"
   B.writeFile fname m
-  pure ""
+  pure (Right ())
 
 fromMail :: [Mailbox]
 fromMail =
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -79,7 +79,7 @@ library
                      , vector >= 0.12.0.0
                      , notmuch >= 0.1 && < 0.2
                      , text
-                     , process
+                     , typed-process >= 0.2.1.0
                      , directory
                      , bytestring
                      , time
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -10,11 +10,13 @@ import Brick.Util (fg, on, bg)
 import qualified Brick.Widgets.Edit as E
 import qualified Graphics.Vty as V
 import qualified Data.ByteString as B
-import qualified Data.ByteString.Char8 as C8
+import qualified Data.ByteString.Lazy as LB
+import qualified Data.Text as T
 import System.Environment (lookupEnv)
 import System.Directory (getHomeDirectory)
 import Data.Maybe (fromMaybe)
-import System.Process (readProcess)
+import System.Process.Typed (proc, readProcessStderr, setStdin, byteStringInput)
+import System.Exit (ExitCode(..))
 
 import Data.MIME (contentTypeTextPlain)
 
@@ -35,18 +37,23 @@ import UI.ComposeEditor.Keybindings
        (listOfAttachmentsKeybindings, composeFromKeybindings,
         composeToKeybindings, composeSubjectKeybindings)
 
+import Error
 import Types
 import Storage.Notmuch (getDatabasePath)
 
 sendmailPath :: FilePath
 sendmailPath = "/usr/sbin/sendmail"
 
--- TODO: see #200
-renderSendMail :: B.ByteString -> IO String
-renderSendMail m =
-        -- -t which extracts recipients from the mail
-        -- TODO: accept a ByteString as stdin, see #201
-        readProcess sendmailPath ["-t", "-v"] (C8.unpack m)
+renderSendMail :: B.ByteString -> IO (Either Error ())
+renderSendMail m = do
+  -- -t which extracts recipients from the mail
+  result <- readProcessStderr config
+  case result of
+    (ExitFailure _, stderr)-> pure $ Left $ SendMailError (decode stderr)
+    (ExitSuccess, _) -> pure $ Right ()
+  where
+    config = setStdin (byteStringInput (LB.fromStrict m)) $ proc sendmailPath ["-t", "-v"]
+    decode = T.unpack . decodeLenient . LB.toStrict
 
 solarizedDark :: Theme
 solarizedDark =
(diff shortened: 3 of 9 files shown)


purebred.cabal
```diff
@@ -79,7 +79,7 @@ library
                      , vector >= 0.12.0.0
                      , notmuch
                      , text
-                     , process
+                     , typed-process
```
frasertweedale: need lower bound >= 0.2.1.0 (when `readProcessStderr` was introduced).

src/Config/Main.hs
```diff
@@ -35,18 +37,24 @@ import UI.ComposeEditor.Keybindings
...
         composeToKeybindings, composeSubjectKeybindings)
 
+import Error
 import Types
 import Storage.Notmuch (getDatabasePath)
 
 sendmailPath :: FilePath
 sendmailPath = "/usr/sbin/sendmail"
 
 -- TODO: see #200
```
frasertweedale: I think this fixes #200 as well, so remove this comment and update commit message.

src/Storage/Notmuch.hs
```diff
@@ -159,8 +160,10 @@ getFromNotmuchConfig :: String -> IO String
 getFromNotmuchConfig key = do
   let cmd = "notmuch"
   let args = ["config", "get", key]
-  stdout <- readProcess cmd args []
-  pure $ filter (/= '\n') stdout
+  (stdout, _) <- readProcess_ $ proc cmd args
```
frasertweedale: non-total (see http://hackage.haskell.org/package/typed-process-0.2.3.0/docs/System-Process-Typed.html#v:readProcess_).

We need to check exit code and handle the failure case (return `Nothing`?)

src/Types.hs
```diff
@@ -201,7 +201,7 @@ data ComposeViewSettings = ComposeViewSettings
     { _cvFromKeybindings :: [Keybinding 'ComposeView 'ComposeFrom]
     , _cvToKeybindings :: [Keybinding 'ComposeView 'ComposeTo]
     , _cvSubjectKeybindings :: [Keybinding 'ComposeView 'ComposeSubject]
-    , _cvSendMailCmd :: B.ByteString -> IO String
+    , _cvSendMailCmd :: B.ByteString -> IO (Either Error ())
```
frasertweedale: :+1:

frasertweedale:
Tests are failing due to now incorrect type of the `writeMailToFile` mail handler.

src/Storage/Notmuch.hs
romanofski: Yes true. I changed it with the equivalent function, since the former function was non-total as well. Will change.

src/Config/Main.hs
romanofski: Ah cheers. Yes!

purebred.cabal
romanofski: Done

src/Storage/Notmuch.hs
romanofski: Hm... so I looked into this and I think it's not worth changing now, because it's a bit of a rabbit hole.

The type for this configuration bit is `IO FilePath` which is converted to a `FilePath` when we process the config. What is missing at this stage is some kind of error handling. Since the function is non-total at the moment and used in the config only, it would basically enforce an error state if the config or notmuch is not correctly setup. Furthermore, it'll show the stderr message.

What happens if I change this to a `IO (Maybe FilePath)` is the fact that I'm deferring the evaluation of the error until pre-processing of the config happens. I could make a little error handler there and maybe exit, but it feels hacky. Alternatively, I could return an empty String which then errors out too, but it'll make a confusing error message. It'd prefer to see the stderr when it fails. 

So perhaps for this patch, make sure stderr is not discarded, but still fail with an exception?

frasertweedale: OK, I'll ACK it for now but would like to follow it up with a commit to either:

- fix `getFromNotmuchConfig` but "handle" the error (i.e. exception if error occurred) in `getDatabasePath` so that type does not change, OR
- absorb `getFromNotmuchConfig` into `getDatabasePath` and leave the behaviour as-is, because it's not used anywhere else.

Got a preference?

romanofski: Hm... how would you handle the error in `getFromNotmuchConfig` so that the type doesn't change? The only thing I can think of is to return an empty string in case of an error, which I don't think will help much?

configs/purebred.hs
```diff
@@ -45,13 +45,13 @@ myMailKeybindings =
 
 -- Note: The empty string return value is currently used as an error condition,
 -- but for this case it is not used.
```
frasertweedale: obsolete comment

src/Storage/Notmuch.hs
frasertweedale: No, the type of `getFromNotmuchConfig` can change (so that it is total); but we move the non-totality (i.e. exception throwing) into `getDatabasePath`, so that *its* type does not change.

configs/purebred.hs
romanofski: :)

src/Storage/Notmuch.hs
frasertweedale: Yep :+1:.  Or just keep using `readProcess_` but move it all into `getDatabasePath` because that's the only place it's called.

romanofski: Alright. Done.

test/TestUserAcceptance.hs
```diff
@@ -691,7 +691,7 @@ precompileConfig testdir = do
   env <- getEnvironment
   let systemEnv = ("PUREBRED_CONFIG_DIR", testdir) : env
       config = setEnv systemEnv $ proc "stack" ["exec", "purebred", "--", "--version"]
-  void $ readProcess_ config
+  void $ runProcess_ config
```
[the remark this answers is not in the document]
romanofski: It prints it like this (except the NixOS part typically):
```
user acceptance tests
  user can view mail:                                       Note: Enabling Nix integration, as it is required under NixOS
Note: Enabling Nix integration, as it is required under NixOS
Launching custom binary /home/rjoost/.cache/purebred/purebred-linux-x86_64

0.1.0.0
OK (3.18s)
    shows tag
    open thread
    view mail
    go back to thread list
    Move down to threaded mails
    Scroll down
    go to next unread mail
    Scroll down (again)
    go to previous mail with reset scroll state
  manipulating notmuch search query results in empty index: Note: Enabling Nix integration, as it is required under NixOS
Note: Enabling Nix integration, as it is required under NixOS
Launching custom binary /home/rjoost/.cache/purebred/purebred-linux-x86_64

0.1.0.0
OK (3.13s)
    focus command
    delete all input
    search for non existing tags yielding no results
    search for mail correctly tagged
    enter new tag
    apply
    open thread
    view currently selected mail
  user can switch back to mail index during composition:    Note: Enabling Nix integration, as it is required under NixOS
Note: Enabling Nix integration, as it is required under NixOS
Launching custom binary /home/rjoost/.cache/purebred/purebred-linux-x86_64
```

stack.yaml
```diff
@@ -58,6 +58,7 @@ extra-deps:
 - brick-0.45
 - megaparsec-6.5.0
 - parser-combinators-1.0.0
+- typed-process-0.2.3.0@sha256:8e9206a2f3c2e1ba2233df97600bbf2bf0007b4670f36b87d456a066d1dee14d
```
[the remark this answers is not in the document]
romanofski: Was suggested by stack and I thought... mh why not. Better remove it?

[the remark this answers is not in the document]
frasertweedale: Nah I don't mind, just wondering what it was.  Digest of the tarball I guess...

Issue #114, opened 2017-12-02
Reply to mail


romanofski:
We need a way to reply to an e-mail. It's currently half implemented without any tests. This looks more trivial than it actually is, since we need to create a quoted body of the replying mail. Also, keep this item to replying to the `from` header only.

frasertweedale:
Or the `Reply-To` header.

romanofski:
Dang! Good point! Didn't think of it...

romanofski:
Since we're using purebred-email now, I think I'll finish this item.

PR #237, opened 2018-12-22
Adds reply to mail feature

romanofski:
This "fixes" the reply to mail action, which was never correctly implemented due
to missing support in the email parsing library. Using purebred-email, there is
wide support for manipulating the mail body in order to properly quote the body.

This is the first take on the replying action and will only work for text/plain
content types.

Fixes https://github.com/purebred-mua/purebred/issues/114

### DIFF
--- a/src/Storage/ParsedMail.hs
+++ b/src/Storage/ParsedMail.hs
@@ -5,10 +5,14 @@
 -- entire mail and it's attachments.
 module Storage.ParsedMail where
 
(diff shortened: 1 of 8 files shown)


frasertweedale:
I haven't used it but I've eyeballed the patch and the UAT steps, all seems fine so ACK when green.  It conflicts with my PR but I can rebase no worries.

PR #239, opened 2018-12-27
Use cabal to run tests

frasertweedale:
Here we go.

The matrix reduces to 5 jobs - one each for 8.0, 8.2, 8.4, 8.6 and head.

Execution time for each job (when dependencies are cached) is around 4..8
minutes depending on worker speed (and the worker's network speed).

Some changes in the UAT suite are not specific to the change to cabal, rather
they are general fixes / robustness improvements.

I've also included the exponential backoff patch again.

Changes:

```
03605b1 (Róman Joost, 2 weeks ago)
   tests: Exponentially back off waiting

   This patch adds an exponential back off to the wait times when waiting for
   a repaint of the UI. With the default of 5 attempts we increase the wait
   time exponentially by 4:

   >>> 20000 / 1e6 0.02
   >>> _ * 4 0.08
   >>> _ * 4 0.32
   >>> _ * 4 1.28
   >>> _ * 4 5.12

   cummulating to 1.2s for the UI to be repainted for every step. This is 
   shorter to the previous setting which waited for every step 0.5s which was
   cumulatively 2.5s until it bailed out.

   This has the potential to shorten the test run time from 98s down to 30s.

   Note: The UI repaint depends on the IO rate of the machine the tests run.
   No consideration has taken into account to determine the IO rate for an
   optimal initial back off time with this patch.

0e22fd8 (Fraser Tweedale, 5 hours ago)
   uat: fix "application has quit" condition

   When the application quits and the shell writes out the command prompt, the
   cursor might not be in column 0.  This could cause the target string to be
   broken across lines, which leads to strange results when ANSI escapes for
   colour/style are involved.

   We can avoid these issues by pressing Enter again, to ensure that the shell
   prompt gets written again from column 0, and by setting PS1 to a relatively
   short string.  But we have another problem - purebred has been sent the 'q'
   key to quit, but if the keys are sent all at once, there's a very good
   chance purebred is still running and eats the keypress events (which
   disappear into the aether).

   To work around these issues:

   - Set PS1 to a string that is otherwise unlikely to appear in the
    purebred output, and

   - After sending 'q', wait a little bit for purebred to terminate,
    then send 'Enter' to write out the prompt again before awaiting
    the expected output

5a03242 (Fraser Tweedale, 4 hours ago)
   uat: make filebrowser tests handle different files

   If for whatever reason the directory in which purebred was run does not
   contain certain files which are currently hardcoded, the tests fail.  This
   can happen if you are running tests from a source distribution instead of a
   git checkout (i.e. the sdist may omit some files).  Make the test more
   resilient by dynamically determining the contents of the directory in which
   purebred is running.

7793357 (Fraser Tweedale, 3 days ago)
   ci: use cabal to build and execute tests

   Use haskell-ci/make_travis_yml_2.hs to regenerate .travis.yml.  It now uses
   cabal to build the program and run tests.

   Some additions to the generated script are required:

   - Extra packages: tmux and notmuch build/run deps.  Travis does not
    merge apt packages lists from top-level and individual matrix
    entries, so we have to fully specify the package list in each
    matrix entry.

   - NOTMUCHVER env var and commands to build and install notmuch

   - Run 'cabal new-install' to install the binary for the UAT suite.
    Include the --symlink-bindir to symlink the executable at
    "${HOME}/.local/bin/".

   - Adjust PATH to include "${HOME}/.local/bin/".

   We also have to handle running the tests from outside the source directory.
    The test suite generated by make_travis_yml_2.hs runs tests from outside
   the source directory.  This confuses the test suite, which uses
   getCurrentDirectory to find some files, and expects purebred to be running
   from the source directory (e.g. for the file browser widget test).

   Enhance the UAT to recognise the "SRCDIR" environment variable. Explicitly
   change directory to the source directory before running purebred.  Update
   .travis.yml to set the SRCDIR environment variable.

0bbf4c2 (Fraser Tweedale, 6 hours ago)
   recompile: add cabal store package-db to ghcOpts

   In case the active GHC package environment does not contain the cabal
   store, determine the path to the newstyle package DB and add a
   -package-db to the GHC options.

4efaa7d (Fraser Tweedale, 33 hours ago)
   tests: include test data in source distribution

   The test files are currently not included in the source distribution. 
   Include them.

   To avoid having to write out a bunch of very long filenames for the 
   contents of Maildir/, make use of wildcard syntax and behaviour available
   since cabal 2.4.  Update the package description accordingly.  This
   entailed moving the "cabal-version" declaration to the top of the file,
   updating the "license" field to include an SPDX identifier (see
   https://spdx.org/licenses/), and adding the
   'autogen-modules' declaration.

   Also drive-by update the "copyright" declaration to include the year 2018.

846aa72 (Fraser Tweedale, 6 days ago)
   uat: run purebred without stack

   When using a cabal new-style install, purebred will be able to run
   (and find the right libraries when recompiling).  Stack-related commands
   and environment variables are not needed, so get rid of them.
```


purebred.cabal
```diff
@@ -20,17 +21,23 @@ description:
...
+-- these wildcard expressions require cabal >= 2.4
+extra-source-files:
+  README.md
+  test/data/Maildir/**/*.url
+  "test/data/Maildir/**/*.url:2,RS"
+  configs/purebred.hs
+
 tested-with:
-  GHC==8.0.2, GHC==8.2.2
+  GHC==8.0.2, GHC==8.2.2, GHC==8.4.4, GHC==8.6.3
```
romanofski: Was wondering if we shouldn't drop 8.0.2? In a way I'd be happy to stick with whatever is currently in Fedora ...

test/TestUserAcceptance.hs
```diff
@@ -152,8 +152,6 @@ testRepliesToMailSuccessfully = withTmuxSession "replies to mail successfully" $
 testFromAddressIsProperlyReset :: TestCase
 testFromAddressIsProperlyReset = withTmuxSession "from address is reset to configured identity" $
   \step -> do
-    setEnvVarInSession "GHC" "stack"
```
romanofski: :+1:

src/Purebred.hs
```diff
@@ -236,14 +236,23 @@ guessPackageArgs dir =
     reversedPath = case reverse path of
       ("lib" : xs) -> xs  -- if component is "lib", drop it
       xs -> xs
-    f s
-      | [".cabal", "store"] `isInfixOf` path
-        && ("purebred-" <> versionString <> "-") `isPrefixOf` s =
-          Just s -- cabal newstyle install
-      | otherwise = Nothing
-    packageId = maybe [] (\s -> ["-package-id", s]) (preview _head reversedPath >>= f)
+    isCabalStore = [".cabal", "store"] `isInfixOf` path
```
romanofski: I hope that doesn't stuff up installing it on NixOS. At the moment I don't have an installable package any way, so lets keep this and fix it in case it does.

test/TestUserAcceptance.hs
```diff
@@ -207,13 +207,25 @@ testUpdatesReadState = withTmuxSession "updates read state for mail and thread"
...
-    sendKeys "q" (Literal "purebred")
+    sendKeys "q" Unconditional
+
+    -- Wait a bit so that purebred, which may not yet have
+    -- terminated, does not eat the upcoming keystroke(s)
+    liftIO $ threadDelay 200000  -- 0.2 seconds
+
+    -- Press Enter again to deal with case where cursor is not at
+    -- column 0, which could cause target string to be split.
+    sendKeys "Enter" (Literal unlikelyString)
```
romanofski: Heh - clever :) Very nice!

romanofski:
Guess merge when green :)

purebred.cabal
frasertweedale: Might as well keep it in the mix until it actually gets in the way of something.

src/Purebred.hs
frasertweedale: Sure, no worries.

PR #240, opened 2019-01-02
Allow to build packages for nix

romanofski:
This adds a few overwrites needed for packages which are not in the known good
set for GHC 8.2.2. Also, we pin Cabal with which we won't be able to build
purebred.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,80 @@
+# Building purebred
+#
+# You should be able to simply invoke:
+#
+# $ nix-build
+#
+# or, to be explicit:
+#
+# $ nix-build default.nix
+#
+# in purebred's root directory in order to build purebred. You'll find the binary under:
+#
+# $ ls result/bin/purebred
+#
+# if the build was successful.
+#
+#
+# Choosing a different compiler than the default
+#
+# In order to choose a different compiler, invoke nix build like so (escaping
+# the quotes is needed, since we're passing a string literal):
+#
+# $ nix-build --arg compiler \"ghc442\"
+#
+#
+# Use as a development environment
+#
+# $ nix-shell default.nix
+#
+{ compiler ? "ghc822" }:
+
+let
+  config = {
+    packageOverrides = super: let self = super.pkgs; in
+    {
+      haskell = super.haskell // {
+        packages = super.haskell.packages // {
+          "${compiler}" = super.haskell.packages."${compiler}".override {
+            overrides = self: super: {
+              # Build with latest known stable version.
+              purebred = with pkgs.haskell.lib; dontHaddock (self.callPackage ./.nix/purebred.nix { });
+              purebred-email = self.callPackage ./.nix/purebred-email.nix { };
+              # Version pins since those versions were bleeding edge and not on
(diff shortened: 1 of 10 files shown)


default.nix
```diff
@@ -0,0 +1,84 @@
...
+let
+  config = {
+    packageOverrides = super: let self = super.pkgs; in
+    {
+      haskell = super.haskell // {
+        packages = super.haskell.packages // {
+          "${compiler}" = super.haskell.packages."${compiler}".override {
+            overrides = self: super: {
+              # Build with latest known stable version.
+              purebred = with pkgs.haskell.lib; dontHaddock (self.callPackage ./.nix-helpers/purebred.nix { });
```
frasertweedale: Why not build the haddock?  Is there a technical reason why we disable it?

default.nix
```diff
@@ -0,0 +1,84 @@
...
+              purebred-email = self.callPackage ./.nix-helpers/purebred-email.nix { };
+              # Version pins since those versions were bleeding edge and not on
+              # hackage/stackage or Nix yet. Remove these once builds without
+              # these pass.
+              notmuch = self.callPackage ./.nix-helpers/hs-notmuch.nix {
+                notmuch = pkgs.notmuch;
```
frasertweedale: Are these two lines (notmuch, talloc) necessary?

default.nix
```diff
@@ -0,0 +1,84 @@
...
+    };
+  };
+  pkgs = import <nixpkgs> { inherit config; };
+  env = pkgs.haskell.packages.${compiler}.ghcWithPackages (self: [
+    self.purebred
+  ]);
+  in
+    if pkgs.lib.inNixShell
+    then pkgs.lib.overrideDerivation pkgs.haskell.packages.${compiler}.purebred.env (old: {
+         buildInputs = old.buildInputs ++ [
```
frasertweedale: Can we clean this up?  Is it necessary?

romanofski: No technical reason. Just took it out to save a few seconds built time. Will remove it.

default.nix
romanofski: No technical reason. Just took it out to save a few seconds built time. Will remove it.

default.nix
romanofski: Ignore the previous comment. Do you mean the commented out line? Yeah I thought I'd keep it as an example, but yeah - happy to remove it too.

default.nix
romanofski: Yes. It specifies which dependency notmuch is coming from (hs-notmuch depends on the C-sources named notmuch). More info I got from: https://github.com/Gabriel439/haskell-nix/blob/9c72b6ecbc5e25df509dfd6ee3d5ee8b9eb21f14/project2/README.md

default.nix
romanofski: Done.

default.nix
romanofski: Done.

frasertweedale:
There are now conflicts with README.  I'm happy for you to merge after resolving.  I think there are a few improvements that can be made but better to merge and I can propose them separately.

PR #244, opened 2019-01-07
explicitly record list lengths in AppState

frasertweedale:
For some lists - namely, those with an underlying type whose
'length' function may not be O(1), and especially if it may force
unwated computation - it is better for efficiency to record the
length separately, or not record it at all.  Add the ListWithLength
type to use for the thread and mail index lists.  Update
'currentItemW' to handle the case where the length is not recorded
(if length is not known it says "Item {i} of ?").

Also drive-by refactor the 'statusbar' function to avoid the
WithContext instance for (Maybe MIMEMessage).

### DIFF
--- a/src/UI/Status/Main.hs
+++ b/src/UI/Status/Main.hs
@@ -7,8 +7,7 @@ import Brick.Types (Widget)
 import Brick.Widgets.Core (hBox, txt, str, withAttr, (<+>), strWrap)
 import qualified Brick.Widgets.List  as L
 import qualified Brick.Widgets.Edit  as E
-import Control.Lens (view)
-import Data.MIME (MIMEMessage)
+import Control.Lens (view, views)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Text.Zipper (cursorPosition)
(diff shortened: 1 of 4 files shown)


src/UI/Status/Main.hs
```diff
@@ -65,11 +61,14 @@ renderStatusbar w s = withAttr statusbarAttr $ hBox
...
   ]
 
-currentItemW :: L.List n e -> Widget n
-currentItemW l = str $
+currentItemW :: ListWithLength t e -> Widget n
+currentItemW (ListWithLength l len) = str $
   maybe
     "No items"
-    (\i -> "Item " <> show (i + 1) <> " of " <> total)
+    (\i -> "Item " <> show (i + 1) <> " of " <> maybe "?" show len)
```
romanofski: Hm... Perhaps in the instance of an unknown length show only the current index instead of a question mark? The question mark doesn't really provide any information but rather looks as if there is a bug.

romanofski:
In an IRC discussion we agreed to merge it for now and worry about the presentation in the status bar later. The plan is to perhaps update the length asynchronously which would show the place holder only for a short period of time perhaps.

PR #246, opened 2019-01-10
implement lazy thread search result loading

frasertweedale:
Here it is!

How to test:

1. Start a big search (e.g. "*")

2. Observe that the search returns quickly, and you are able to scroll around.
   But do not jump to bottom!

3. Observe that the number of items is reported as "?"

4. Watch in breathless awe as, at some point, the "?" turns into a number!

5. Jump to bottom.  This should, now that the spine of the list has been
   traversed, in order to determine its length, be near-instantaneous.

For code review, it is best to digest this PR on a commit-by-commit basis.

Changes (reverse order):

```
20e58ed (Fraser Tweedale, 3 days ago)
   use correct notmuch SearchTerm constructors

   The 'FreeForm' constructor is not correct.  Although the current
   'show' instance does not quote it, this is a bug.  The correct constructor
   for passing a query string to libnotmuch as-is is Bare. Change to the
   correct contstructor and also access them via the qualified Notmuch import
   (Notmuch.Search will be removed in a future release).

b4548db (Fraser Tweedale, 18 minutes ago)
   move event types to new module

   Add the Purebred.Events module and move types and functions related to
   events to it.

   Update the generation 'Generation' to hide the constructor and avoid the
   'Enum' instance.  This avoids a conflict between
   -XDeriveAnyClass and -XGeneralizedNewtypeDeriving, which caused a build
   failure with GHC 8.0 and warnings with later versions.  The removal of the
   'Enum' type class also lets us enforce monotonicity.

bb88515 (Fraser Tweedale, 21 hours ago)
   ignore obsolete NotifyNumThreads events

   If a search with a small result is performed after a search with a large
   result, the NotifyNumThreads event for the earlier search can arrive after
   the NotifyNumThreads event for the later search.  As a consequence, the
   wrong number of items will be displayed.

   Define the 'Generation' data type for recording a particular
   "generation" of the 'AppState' (or some aspect thereof).  Events can also
   carry a 'Generation' to indicate the generation(s) of the AppState to which
   it is related.

   Update the 'NotifyNumThreads' event data type and the application state for
   the thread index to each include a generation field. Update the event
   handler to ignore 'NotifyNumThreads' events that do not match the current
   generation of the thread index.

   Part of: https://github.com/purebred-mua/purebred/issues/233

367a739 (Fraser Tweedale, 2 days ago)
   fork thread to compute length of thread index

   After performing a notmuch search, fork a thread to compute the length of
   the result (fully evaluating the spine of the list) and send a notification
   to brick to update the model.

   This adds the 'NotifyNumThreads' constructor to the 'PurebredEvent' type,
   and updates the event handler to process it.

   The 'initialState' function is refactored to reuse 'applyThreads', where
   the "fork background thread" behaviour is defined, so that the thread is
   spawned for the initial search upon startup, too.

   Part of: https://github.com/purebred-mua/purebred/issues/233

9c8c21f (Fraser Tweedale, 5 weeks ago)
   lazily load threads

   If search result is very long, it takes a long time to load all the items. 
   Preliminary work in brick, hs-notmuch and purebred leads to what is now a
   small change to enable lazy loading of search results.

   First, we move to (Haskell) notmuch-0.2 which fixes issues with lazy result
   loads.  In particular, derived objects always hold a reference to their
   parent object, to prevent it being GC'd, which caused segfaults when the
   child object (wihch is an opaque pointer) held pointers into the parent. 
   One programmer-friendly consequence of this is that resource management of
   the database handle is now implicit.  The database will be closed when it
   gets GC'd, which will only happen when all derived objects (queries,
   threads and messages) have been GC'd.

   Second, we update the thread index list to use the 'V' container type added
   a little while ago, which supports lazy loading.

   Third, for processing notmuch Thread objects into purebred's NotmuchThread
   type, we define a []- and IO-specific 'lazyTraverse' function that uses
   'unsafeInterleaveIO' to retain the lazy loading behaviour afforded by
   hs-notmuch.

   Finally, we tweak the functions that update the AppState to not set the
   'length' of the list (because 'length' would force the entire spine of the
   container).

   Fixes: https://github.com/purebred-mua/purebred/issues/233
```

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -719,7 +719,7 @@ notifyNumThreads :: (MonadIO m, Foldable t) => AppState -> t a -> m AppState
 notifyNumThreads s l =
   let
     len = length l
-    nextGen = views (asMailIndex . miListOfThreadsGeneration) succ s
+    nextGen = views (asMailIndex . miListOfThreadsGeneration) nextGeneration s
     s' = set (asMailIndex . miListOfThreadsGeneration) nextGen s
     go = len `seq` writeBChan (view (asConfig . confBChan) s) (NotifyNumThreads len nextGen)
   in
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -703,7 +703,7 @@ applySearch s = runExceptT (Notmuch.getThreads searchterms (view (asConfig . con
    where searchterms = currentLine $ view (asMailIndex . miSearchThreadsEditor . E.editContentsL) s
          updateList vec =
            over (asMailIndex . miThreads . listList) (L.listReplace vec (Just 0))
-           . set (asMailIndex . miThreads . listLength) (Just (length vec))
+           . set (asMailIndex . miThreads . listLength) Nothing
 
 setMailsForThread :: AppState -> IO AppState
 setMailsForThread s = selectedItemHelper (asMailIndex . miListOfThreads) s $ \t ->
@@ -714,10 +714,10 @@ setMailsForThread s = selectedItemHelper (asMailIndex . miListOfThreads) s $ \t 
   in either setError updateThreadMails <$> runExceptT (Notmuch.getThreadMessages dbpath t)
 
 selectedItemHelper
-    :: Applicative f
-    => Getting (L.List n t) AppState (L.List n t)
+    :: (Applicative f, Foldable t, L.Splittable t)
+    => Getting (L.GenericList n t a) AppState (L.GenericList n t a)
     -> AppState
-    -> (t -> f (AppState -> AppState))
+    -> (a -> f (AppState -> AppState))
     -> f AppState
 selectedItemHelper l s func =
   ($ s) <$> case L.listSelectedElement (view l s) of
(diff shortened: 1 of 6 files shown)


src/UI/Actions.hs
```diff
@@ -697,14 +700,25 @@ listSetSelectionEnd
...
+    Right threads -> notifyNumThreads s threads $> updateList threads s
    where searchterms = currentLine $ view (asMailIndex . miSearchThreadsEditor . E.editContentsL) s
          updateList vec =
            over (asMailIndex . miThreads . listList) (L.listReplace vec (Just 0))
            . set (asMailIndex . miThreads . listLength) Nothing
 
+-- | Fork a thread to compute the length of the container and send a
+-- NotifyNumThreads event.  'seq' ensures that the work is actually
+-- done by the spawned thread.
+notifyNumThreads :: (MonadIO m, Foldable t) => AppState -> t a -> m ()
```
romanofski: Would we gain better control/robustness if we'd use the async package? Guess the current use doesn't strike me as too complex, so perhaps it's something for later.

PR #247, opened 2019-01-10
Adjust the notmuch version for nix builds

romanofski:
Because the latest notmuch version bump, we'll need to bump the nix build
version too.


frasertweedale:
@romanofski anything that only touches nix, feel free to just push.

PR #250, opened 2019-02-02
Nix improvements

romanofski:


romanofski:
Don't merge yet. Found a problem with the notmuch binding.

romanofski:
I think it's ready to merge. I wish I could smuggle in a nix-build step in Travis ....

romanofski:
Turned out, making use of Travis to build with nix was easier than I thought.

romanofski:
This is even more interesting than I thought. I guess I won't be able to include a nix test overwriting the compiler to use GHC-8.4.4, since I don't know a way to overwrite cabal in order to use >= 2.4. Instead I overwrite the compiler and use GHC HEAD. Not sure if that is any better tho.

test/TestUserAcceptance.hs
```diff
@@ -939,13 +937,15 @@ sendLiteralKeys keys = do
     waitForString keys defaultCountdown
 
 capture :: ReaderT Env IO String
-capture =
-  tmuxSessionProc "capture-pane"
+capture = T.unpack . decodeLenient . LB.toStrict
```
frasertweedale: Just as an FYI, `text` provides functions for decoding lazy bytestrings directly.
I don't think we need bother changing to use it here though (i.e. this shouldn't
be a performance issue here, and even if it was it's in the test suite so I don't
care ^_^)

frasertweedale:
Fine with me.  Merge ho!

test/TestUserAcceptance.hs
romanofski: Ah cheers! Good to know

PR #252, opened 2019-02-06
tests: use nix-shell

romanofski:


romanofski:
This now runs builds and runs the uat in the nix shell. I think that should prove that we can not regress on that part. I think for the problem that the filepath is too long, I'll create a separate issue and ignore that for now. If the test fails all the time I guess I'll adjust the test or... mh.. maybe fix the UI with a quick fix... not sure.

romanofski:
... and there I thought right aligning the filepath will help, but unfortunately shows the same problem. Hrm. I'll need to think of a better solution.

I'm thinking of trying out to visually shrink the path. We're constructing the full path to the attachment by a concatenation of `search path + user selected path`. If, for the time being, when rendering the path I replace the `search path` part by a `~` I might be able to get away with it. Would still need a proper fix tho, since `~` is not always set as the search path.

romanofski:
Alright. I think only showing the filename should do the trick. I think that's currently the best workaround, because
a) the e-mail gets serialised to using only the filename anyway
b) it's only presented as only the filename. The full path is still kept.
c) defer a better solution to be implemented in #253 
d) my initial idea using `~` I think is more confusing

frasertweedale:
All fine with me. Merging.

PR #254, opened 2019-02-15
Refactor/181 with vector

romanofski:


romanofski:
... odd - all builds on the branch are passed. Seems like the message got lost somewhere ...

frasertweedale:
Subsumed by #257

PR #257, opened 2019-02-23
Feature/manage attachments

romanofski:
Here we go: listing, opening and piping attachments. Thankfully the default for piping shows you the output so there is not much more work to be done.

This is based upon #254 which will have to be reviewed first (or.. review this and throw #254 away).

This is all a bit primitive and I suppose subsequent work could make all of this way more pluggable and more safe. Yet it shall do for getting by.

Also, I think there will be a few backlog items for future work I could think of ...


romanofski:
Some changes:

* use `byteStringInput` for the `pipeCommand`
* remove the temporary file after command exited
* handle arguments of the typed in commands (use `less -e` for the test)

PR #266, opened 2019-03-26
constrain list actions to contexts with lists

frasertweedale:
Just a refactor I've wanted to do for a little while...

Changes:

```
a31b581 (Fraser Tweedale, 76 minutes ago)
   add purebred-email upper bound

   Ahead of anticipated breaking API changes in purebred-email, constrain
   purebred to <= 0.2.

deadd88 (Fraser Tweedale, 76 minutes ago)
   remove obsolete language extension

425c967 (Fraser Tweedale, 2 hours ago)
   constrain list actions to contexts with lists

   Add the HasList class and constrain list actions to contexts that have an
   instance of HasList.  This simplifies the implementation of current (or
   future) list actions.

   Also fix descriptions of the list actions.
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
 module Main where
 
 import Purebred (purebred, defaultConfig)
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -145,7 +145,7 @@ test-suite uat
   main-is:             TestUserAcceptance.hs
   default-language:    Haskell2010
   build-depends:       base
-                     , purebred-email
+                     , purebred-email >= 0.1 && < 0.2
                      , tasty
                      , tasty-hunit
                      , directory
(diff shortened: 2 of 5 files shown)


frasertweedale:
Guess I messed something up...

app/Main.hs
```diff
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
```
romanofski: How do you figure out that some are not in use? Just by knowing? hlint isn't telling you, right?

frasertweedale:
Why yes, it would @romanofski.  Thanks for pointing that out.

I can't repro the UAT failure when running the test.  Do you have any comment?

app/Main.hs
frasertweedale: hlint told me.  But also, this file is tiny and it's easy to see there is no string literal, let alone an overloaded one :)

romanofski:
Let me have a look ...

frasertweedale:
> Let me have a look ...

nevermind, I was only running the unit tests... :/

frasertweedale:
@romanofski I have Questions.  I'll need to discuss this with you online I think.  It can wait.

PR #274, opened 2019-03-29
Update the goal list

romanofski:
This adds two more points of goals we want to achieve: extensive testing and
better security through the use of Haskells type system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ are welcome.  See [HACKING] for more info.
 - [ ] re-use strong UX patterns from existing console mailers like mutt and sup
 - [ ] extensible with plug-ins
 - [x] enforce sane configuration by using Haskell's type system
+- [ ] enforce security against malicious emails by using more expressive types
+- [x] provide a reliable application using a large suite of user acceptance tests
 
 ## Try it
 


README.md
```diff
@@ -23,6 +23,8 @@ are welcome.  See [HACKING] for more info.
 - [ ] re-use strong UX patterns from existing console mailers like mutt and sup
 - [ ] extensible with plug-ins
 - [x] enforce sane configuration by using Haskell's type system
+- [ ] enforce security against malicious emails by using more expressive types
```
romanofski: Didn't tick it yet, since I thought there is more to come (I had your recent commits of the tainting of data in mind)? What do you think?

frasertweedale: Sure.  I just think maybe we want to move the whole section a bit further down in the README, because they look like huge gaps, even if we got to 90% of where we want to be...

romanofski: Hm... maybe it's not the right format then? Perhaps changing it into growing feature list which puts us apart and points out the motivation as to why we're implementing a new mailer would be better?

frasertweedale: I dunno, I'll just merge this, we can re-org later.

PR #275, opened 2019-03-30
Mailcap to automatically open attachments

romanofski:
This adds a proof of concept way to lookup commands for content types. Some notes I'm currently thinking over:
* The new `Action` encapsulates two actions with a condition (can we lookup a content type for which a command is defined?) and continues from that. We currently have no ability to access the `AppState` from our `Keybinding`s. Perhaps we don't need to and keep those `Action`s "special". I would keep it as it is and perhaps create a new backlog item for the future. We already have a few more items like this (e.g. `invokeEditor`)
* the mailcap is a `Map.Map (CI ByteString, CI ByteString) String`. The lookup for the content type is a bit clumsy at the moment. I figured that we most likely don't want to specify the full `ContentType` since we don't need different commands for different `Parameter`s. At least I couldn't think of a good reason. I'm not sure if I leave the key as a tuple or make a better data type.

### DIFF
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -17,7 +17,7 @@ import System.Directory (getHomeDirectory)
 import Data.Maybe (fromMaybe)
 import System.Exit (ExitCode(..))
 
-import Data.MIME (contentTypeTextPlain)
+import Data.MIME (contentTypeTextPlain, matchContentType)
 
 import UI.FileBrowser.Keybindings
        (fileBrowserKeybindings, manageSearchPathKeybindings)
@@ -178,6 +178,10 @@ defaultConfig =
       , _mvMailListOfAttachmentsKeybindings = mailAttachmentsKeybindings
       , _mvOpenWithKeybindings = openWithKeybindings
       , _mvPipeToKeybindings = pipeToKeybindings
+      , _mvMailcap = [
+            ((matchContentType "text" (Just "html")), "elinks -force-html")
+          , (const True, "xdg-open")
+          ]
       }
     , _confIndexView = IndexViewSettings
       { _ivBrowseThreadsKeybindings = browseThreadsKeybindings
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -347,6 +347,7 @@ data MailViewSettings = MailViewSettings
     , _mvMailListOfAttachmentsKeybindings :: [Keybinding 'ViewMail 'MailListOfAttachments]
     , _mvOpenWithKeybindings :: [Keybinding 'ViewMail 'MailAttachmentOpenWithEditor]
     , _mvPipeToKeybindings :: [Keybinding 'ViewMail 'MailAttachmentPipeToEditor]
+    , _mvMailcap :: [(ContentType -> Bool, String)]
     }
     deriving (Generic, NFData)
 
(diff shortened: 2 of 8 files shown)


romanofski:
Hm.. I've googled around a little bit and perhaps declaring our own mailcap file is not the way to go anymore. Reason being: https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html That integrates well within the desktop environment and has a fallback for the console.
The disadvantage is the additional burden to learn new tools, but coming from freedesktop.org I feel it's perhaps the better way to go?

frasertweedale:
@romanofski well, mailcap is somewhat standardised too.  They are not mutually exclusive.  I envisage we would have our own types for specifying how a type should be handled, and then we can have plugins to populate that data (or look it up on the fly) in mailcap / XDG configuration / whatever.  My $0.02.

src/Config/Main.hs
```diff
@@ -178,6 +179,7 @@ defaultConfig =
       , _mvMailListOfAttachmentsKeybindings = mailAttachmentsKeybindings
       , _mvOpenWithKeybindings = openWithKeybindings
       , _mvPipeToKeybindings = pipeToKeybindings
+      , _mvMailcap = Map.fromList [(("text", "html"), "elinks -force-html")]
```
frasertweedale: If we want to reflect what mailcap is capable of, we need to support "any" (i.e. wildcard) for the subtype.  To mirror what *purebred-email* does we would have `(CI B.ByteString, Maybe (CI B.ByteString))`.  But really we could just use the type [(ContentType -> Bool, String {- comand -}], and traverse this list until we get a hit.  That means we could use `Data.MIME.matchContentType` to easily construct the predicate function:
```haskell
matchContentType
  :: CI B.ByteString         -- ^ type
  -> Maybe (CI B.ByteString) -- ^ optional subtype
  -> ContentType
  -> Bool
```

romanofski: Yeah true. I hoped with the Map it would have been easier, but really there isn't any benefit at the moment. Cheers

romanofski: Alright updated. However, I didn't go with the:
```Haskell
[(ContentType -> Bool, String {- comand -}]
```
but used instead:
```Haskell
[(CI B.ByteString, Maybe (CI B.ByteString))]
```
Reason being that I can't see the point of taking a function with me for the first part of the tuple. Misunderstanding maybe?

romanofski: Just realized what you meant. Hm happy to update the patch or merge of it's not a deal breaker.

frasertweedale: Don't merge yet, I'd like to play with it, but yeah `[(type, Maybe subtype)]` is fine with me for now.

romanofski:
I thought about it over night and using a function would perhaps help to allow adding a bare rule e.g.: 
```Haskell
[(const True, "xdg-open")]
```
Not sure if that makes sense to pass everything to `xdg-open` from a security perspective, but well it would be a possibility. On the other hand it might be better to change it if the need arises ...

frasertweedale:
@romanofski yes, you've hit on one of the other benefits of using a function :)

romanofski:
Let's go with the function then. I'll update the patch.

PR #277, opened 2019-03-30
uat: ignore test return type

frasertweedale:
The 'pure ()' at the end of almost every test is an annoyance we can
avoid.  Avoid it via void.


romanofski:
Ah yes of course!

Issue #278, opened 2019-04-03
year not shown on old mails


frasertweedale:
**Describe the bug**
When reading an old mail (or in the thread list) the actual date/time/year of the mail is not shown.

**Expected behavior**

- Full date/time information should be shown in default headers.
- For threads > 1y old, year should be shown in date field of thread list.

romanofski:
Yeah true. Not that there is room for it.

frasertweedale:
@romanofski maybe `DD Mon` for mail < 1y old, and `Mon YY` for > 1y?

romanofski:
Yeah sounds fine. I wonder if it's alright to not align it, so that an addional year pushes the rest to the right.

frasertweedale:
How do you mean?  `DD Mon` and `Mon YY` are the same width anyway...

romanofski:
Ah darn. Yes indeed! Never mind!

PR #282, opened 2019-04-03
Show the calendar year if mails are older than a year

romanofski:
This add a change in the date format shown in the inbox: if the mail is older
than a year from today, the date is formatted as Month/Year instead of
Day/Month.

Fixes https://github.com/purebred-mua/purebred/issues/278


romanofski:
I wanted an easy fix. I tried it and .. I don't know.. it's a bit odd when it switches. You think: what the heck? Did the day just move over next to the month? Would you mind trying it out?
PS: Misses one other item: showing the date header in the mail so don't merge yet.

romanofski:
Alright I played around with it some more and adjusted it a bit which I think makes I easier to understand the difference. For example, less than one year:
```
# recent
01/Apr
# older with a white space
Sep 18
```

frasertweedale:
@romanofski pushed another commit to hopefully fix the faily ci on GHC 8.0.  Needed to add a lower bound on 'time'.

frasertweedale:
Uh, it's the test that constrains to installed instances, and time-1.6 is installed.  Maybe we should just drop 8.0 (all OSen I care about are at 8.2 or later).  What say you @romanofski ?

romanofski:
Yeah lets drop 8.0... It's not that we have some crazy support contract to maintain that much backwards compatibility.

frasertweedale:
I might suggest one tiny tweak: an apostrophe between month and year for >1y.
```
01/Apr
Sep'18
```

src/UI/App.hs
```diff
@@ -147,8 +148,9 @@ initialState conf =
          (E.editor ManageFileBrowserSearchPath Nothing path)
     mailboxes = view (confComposeView . cvIdentities) conf
     s = AppState conf mi mv (initialCompose mailboxes) Nothing viewsettings fb
-  in
-    applySearch s
+  in do
+    t <- zonedTimeToUTC <$> getZonedTime
```
frasertweedale: so, we only set the date once?  Shouldn't we update it from the current system time each time we run a search (or alternatively, set a timer to do it?)

src/Types.hs
```diff
@@ -473,6 +473,7 @@ data AppState = AppState
     , _asError     :: Maybe Error -- ^ in case of errors, show this error message
     , _asViews     :: ViewSettings -- ^ stores widget and focus information
     , _asFileBrowser :: FileBrowser
+    , _asLocalTime :: UTCTime
```
frasertweedale: Is it possible to avoid putting this in the state at all, and ask the current time in the drawing function?

src/UI/App.hs
romanofski: Yeah actually you have a point. I had in mind that you eventually close the application and start it again and being off by a few days wouldn't hurt. But perhaps using a timer is the better choice.

src/Types.hs
romanofski: We can't do IO in the drawing code, but we could in the event handling. Hmmm I think keeping it as a state variable is the only choice here...

frasertweedale: I think you're right.

src/UI/App.hs
frasertweedale: Do the simple thing first IMO - update the time each time we do a new notmuch search.

src/UI/Actions.hs
```diff
@@ -844,9 +845,10 @@ updateBrowseFileContents contents s =
 applySearch :: (MonadIO m) => AppState -> m AppState
 applySearch s = do
   r <- runExceptT (Notmuch.getThreads searchterms (view (asConfig . confNotmuch) s))
+  t <- liftIO (zonedTimeToUTC <$> getZonedTime)
   case r of
     Left e -> pure $ setError e s
-    Right threads -> updateList threads <$> notifyNumThreads s threads
+    Right threads -> updateList threads <$> notifyNumThreads s threads >>= pure . set asLocalTime t
```
frasertweedale:
```haskell
Right threads -> set asLocalTime t . updateList threads <$> notifyNumThreads s threads
```

`(>>= pure . f)` can always be replaced with a simple `fmap f` :)

romanofski: Done. Had some trouble with the `fmap` but hopefully got it right.

src/UI/Actions.hs
```diff
@@ -844,9 +845,11 @@ updateBrowseFileContents contents s =
 applySearch :: (MonadIO m) => AppState -> m AppState
 applySearch s = do
   r <- runExceptT (Notmuch.getThreads searchterms (view (asConfig . confNotmuch) s))
+  t <- liftIO (zonedTimeToUTC <$> getZonedTime)
   case r of
     Left e -> pure $ setError e s
-    Right threads -> updateList threads <$> notifyNumThreads s threads
+    Right threads -> set asLocalTime t
+                     <$> (updateList threads <$> notifyNumThreads s threads)
```
frasertweedale: surely `set asLocalTime t . updateList threads <$> notifyNumThreads s threads` ... ?

src/UI/App.hs
```diff
@@ -146,9 +148,9 @@ initialState conf =
          (L.list ListOfFiles mempty 1)
          (E.editor ManageFileBrowserSearchPath Nothing path)
     mailboxes = view (confComposeView . cvIdentities) conf
-    s = AppState conf mi mv (initialCompose mailboxes) Nothing viewsettings fb
-  in
-    applySearch s
+    epoch = UTCTime (fromGregorian 2018 07 18) 1
```
frasertweedale: Heh, what epoch is this?  (Not suggesting to change it... just curious :) )

romanofski: First commit :)

src/UI/Actions.hs
romanofski: Sometimes my representation of `fmap` as being something I can loop over containers in an abstract way other than a general abstraction keeps me from doing the right thing here. I read it as set the local time for each thread instead of a better `>>=`... ah well

src/UI/App.hs
frasertweedale: Off by one year error :D  It was 2017.  (Don't worry about it).

PR #285, opened 2019-04-20
Fix/276

romanofski:


src/UI/Actions.hs
```diff
@@ -585,11 +579,14 @@ openAttachment =
...
-    let cmd = view (asMailView . mvOpenCommand . E.editContentsL . to (T.unpack . currentLine)) s
-    in Brick.suspendAndResume $ liftIO $ openCommand' s cmd
-  }
+    { _aDescription = ["ask for command to open attachment"]
+    , _aAction =
+        \s ->
+          let cmd = view (asMailView . mvOpenCommand . E.editContentsL . to (T.unpack . currentLine)) s
+           in if null cmd
+                then Brick.continue $ setError (GenericError "Empty command") s
+                else Brick.suspendAndResume $ liftIO $ openCommand' s (MailcapHandler (Process (fromList cmd) []) False)
```
frasertweedale: Rather than use `null` and the non-total `fromList`, instead:
```haskell
case cmd of
  [] -> error case
  x:xs -> blah blah (Process (x :| xs)) blah
```

src/Config/Main.hs
```diff
@@ -15,6 +15,7 @@ import qualified Data.Text as T
 import System.Environment (lookupEnv)
 import System.Directory (getHomeDirectory)
 import Data.Maybe (fromMaybe)
+import GHC.Exts (fromList)
```
frasertweedale: Can we import `Data.List.NonEmpty.fromList` instead?

src/Purebred/System/Process.hs
```diff
@@ -33,8 +37,12 @@ import System.Exit (ExitCode(..))
 import Control.Exception (try, IOException)
 import System.Process.Typed
 import qualified Data.ByteString.Lazy as LB
+import qualified Data.ByteString as B
+import System.IO.Temp (emptySystemTempFile)
+import System.Directory (removeFile)
 import Control.Lens (set, (&))
 import Data.Semigroup ((<>))
+import GHC.Exts (toList)
```
frasertweedale: Can we use `Data.Foldable.toList` instead?

src/UI/Actions.hs
```diff
@@ -498,7 +493,7 @@ class ViewTransition (v :: ViewName) (v' :: ViewName) where
 instance ViewTransition v v where
 
 instance ViewTransition 'Mails 'Threads where
-  transitionHook _ _ = set (asViews . vsViews . at Mails . _Just) listOfMailsView
+  transitionHook _ _ = set (asViews . vsViews . ix Mails) listOfMailsView
```
frasertweedale: Wow, it is nice that HLint groks some lens stuff :)

src/UI/Actions.hs
romanofski: True!

src/Config/Main.hs
romanofski: Yep

src/UI/Actions.hs
romanofski: Yeah hey. I like it.

romanofski:
Can I merge this?

frasertweedale:
@romanofski sorry I just reviewed the updates then, and merged it :)

PR #287, opened 2019-04-24
Feature/colored tags

romanofski:


configs/purebred.hs
```diff
@@ -67,3 +67,12 @@ main = purebred $ tweak defaultConfig where
...
     . set (confComposeView . cvIdentities) fromMail
+    . over confTheme (applyAttrMappings myColoredTags)
+
+myColoredTags :: [(AttrName, Attr)]
+myColoredTags =
+  [ ("inbox", fg cyan)
+  , ("archive", fg cyan)
+  , ("signed", fg green)
+  , ("attachment", fg brightRed)
+  ]
```
frasertweedale: what happens if the tag does not match any of these?  What colour/style does it get?

IMO we want to construct the `AttrName` for the tag name as something like `"tag" <> tagName` where `tagName` is `"inbox"`, `"signed"`, etc.  This would allow to specify the default style for a tag name (with attr name `"tag"`) as well as overriding the style for specific tags.

src/UI/Index/Main.hs
```diff
@@ -126,4 +125,9 @@ renderAuthors isSelected authors =
 renderTagsWidget :: [Tag] -> Tag -> Widget Name
 renderTagsWidget tgs ignored =
     let ts = filter (/= ignored) tgs
-    in withAttr mailTagsAttr $ vLimit 1 $ txt $ decodeLenient $ B.intercalate " " $ fmap getTag ts
+        render tag = padRight (Pad 1) $ withAttr (toAttrName tag) $ txt (decodeLenient $ getTag tag)
+    in vLimit 1 $ hBox $ render  <$> ts
+
+toAttrName :: Tag -> AttrName
+toAttrName t = let attr = attrName . unpack . decodeLenient . getTag
+               in attr t
```
frasertweedale: We can clean this up:

```haskell
toAttrName = attrName . unpack . decodeLenient . getTag
```

And, to elaborate what I suggested above, this would become like:

```haskell
toAttrName = ("tag" <>) . attrName . unpack . decodeLenient . getTag
```

romanofski: AAah yes I see what you mean. Will try.

configs/purebred.hs
romanofski: The default is returned. I need to specify it when I create the attribute map. In `Config.Main`:
```Haskell
solarizedDark :: A.AttrMap
solarizedDark =
    A.attrMap
        V.defAttr  -- <-- default
        [ (listAttr, V.brightBlue `on` V.brightBlack)
--  and so on
```

src/UI/Index/Main.hs
romanofski: Hm... yeh.. your idea of returning the `"tag"` default style won't work with our current setup. The function `toAttrName` will always return a tag which means you always return `"tag"` with `"something"`.

If you wanted to support a fallback just for `"tag"`, we'd need a separate `AttrMap` and set the default attribute of that map accordingly. At the moment it's basically a global map. At this point, if you want colored tags, just use the what's in the default config and you're good to go.

Having said that, I'd still keep your suggestions.

frasertweedale: Right, but can't we change the config to refer to `"tag" <> "foo"`, and it will work as expected?

frasertweedale: I haven't tried anything with attrmaps yet,  so maybe my intuition / interpretation of the docs is wrong...

configs/purebred.hs
frasertweedale: Yeah but isn't that a default style for *everything*, not just tags?

src/UI/Index/Main.hs
romanofski: Hm not sure what you mean by

> change the config to refer to `"tag" <> "foo"`

The current map is global and stores attributes for the entire application. If you lookup an attribute which isn't in the map, it's rendered with the default attribute, currently `defAttr` from `Vty`. That happens whether we use it for tags or whatever. If we change the `defAttr` to be the color definition for a default Tag than any other undefined element would use the same color definition as the default tag.

The other alternative I could imagine is to query the `AttrMap` if `"tag" <> "foo"` is in it and if not return the default...

frasertweedale: AttrMap lookup is heirarchical.  i.e. if you look for `"foo"<>"bar"` in a map that only has `"foo"`, then `"foo"` style will apply.

So you can have a style for `"tag"`, which is the default for all attrs beginning with `"tag"`, and then override it for more specific tags (e.g. `"tag"<>"inbox"`).

At least, that is my interpretation of the docs: https://hackage.haskell.org/package/brick-0.47/docs/Brick-AttrMap.html#v:attrMapLookup

configs/purebred.hs
romanofski: Yes. That is correct.

romanofski:
What to do:

* merge
* wait and add a separate attribute map for tags with a sensible default
* wait and add code to inspect the global attribute map and fall back to the default attribute
* wait and do more investigation (not sure into what)

I guess the second point would give a sensible default with the cost of another map. Third would add more code without another map. Both give a more pleasant fallback when it comes to colours.

frasertweedale:
@romanofski I'm happy with it.  Merge ho!

PR #291, opened 2019-05-04
Add stack.yaml

gnull:
For some reason I couldn't install nix on my Arch linux machine. So to compile this project I manually crafted this `stack.yaml` file, which worked for me and I have successfully build this project using it.

I'm not sure if you want to have a `stack.yaml` file here, so opening a pull request just in case. Hope you find it useful!


romanofski:
Cheers @gnull for your contribution! 
The topic on stack is actually a bit more complicated. We actually had started off using stack, but since @frasertweedale is an avid cabal user while I'm trying out NixOS we removed stack in 846aa724df2074e34128b402e8bc319340c689a7 and ce62409f2a618e608f0cc713ab1a76eac6960bff. It goes a little bit beyond just usage preferences, since we couldn't make 846aa724df2074e34128b402e8bc319340c689a7 to work with stack.

I do know that it is tricky to get around deciding which build tools to choose for Haskell. The following are alternative solutions I could think of moving forward:
1. You're just trying out purebred and want to leave it at that. We keep what we have and leave stack out for now.
2. You're interested in hacking on purebred and want to use stack. In order to make sure that the stack support doesn't regress we would need:
    1. An addition to the `.travis.yml` job which runs a `stack build` similar to what we currently have with `nix-build`.
    2. A way we can support the re-compilation of the `purebred.hs` configuration file which will probably have to invoke `stack`.
3. Our documentation is lacking a "How to get started with cabal" section and you would want to give `cabal` a go. I'm happy to add a section in our `HACKING` guide.

We'd be happy to help were we can. Let us know :)

gnull:
@romanofski, thank you for such a warm welcome!

I think the best option for now is to leave the stack out.

The purebred is an interesting project, it looks very promising to me. I might get involved in hacking on it later, but for now I'll just watch it closely.

Ivan

PR #292, opened 2019-05-10
Use most dependencies from within nix

romanofski:
Remove our custom nix derivations and use the derivations out of the current nix
release instead.


romanofski:
Hmpf... for whatever reason it picks too many packages which seem to not be in the cache. I'll need to find out why ...

romanofski:
I'll close this for now. Back to the drawing board.

PR #293, opened 2019-05-19
Refactor IO error handling for attachments

romanofski:
With the refactoring for handling temporary files (see
dd8ad0d062a418c3bc09ef63c6de36b5b5af908a) we introduced a slightly
different way to handle errors. This is slightly different to the way
we've handled errors from IO computations before.

Throughout most of our code, we use MonadError in order to transform and
propagate errors in IO computations.

This patch changes the error reporting we used to be consistent to what
we have throughout the code base. Most types have changed to use
MonadError as well as functions in order to propagate the error to the
user interface.

This becomes especially important when we further make changes to the IO
code which needs extensive error handling and propagation to the user
interface.

### DIFF
--- a/src/Purebred/System/Process.hs
+++ b/src/Purebred/System/Process.hs
@@ -13,16 +13,18 @@
 --
 -- You should have received a copy of the GNU Affero General Public License
 -- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+{-# LANGUAGE FlexibleContexts #-}
 
 module Purebred.System.Process
-  ( tryRunProcess
+  ( tryReadProcess
   , handleIOException
   , handleExitCode
   , Purebred.System.Process.readProcess
   , tmpfileResource
   , emptyResource
   , toProcessConfigWithTempfile
-
+  , runEntityCommand
+  , tryIO
   -- * Re-exports from @System.Process.Typed@
   , ProcessConfig
   , proc
@@ -32,18 +34,21 @@ module Purebred.System.Process
   ) where
 
 import Data.Bifunctor (bimap)
+import Data.Functor (($>))
 import System.Exit (ExitCode(..))
 import Control.Exception (try, IOException)
+import Control.Monad.Catch (bracket, MonadMask)
+import Control.Monad.IO.Class (liftIO, MonadIO)
 import System.Process.Typed
+import Control.Monad.Except (MonadError, throwError)
 import qualified Data.ByteString.Lazy as LB
 import qualified Data.ByteString as B
 import System.IO.Temp (emptySystemTempFile)
 import System.Directory (removeFile)
-import Control.Lens (set, (&))
+import Control.Lens ((&), _2, over, set, view)
 import Data.Semigroup ((<>))
 import Data.Foldable (toList)
 
-import Control.Monad.IO.Class (MonadIO)
 import qualified Data.Text as T
 
 import Error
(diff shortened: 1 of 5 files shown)


src/UI/Actions.hs
```diff
@@ -1052,7 +1054,10 @@ openCommand' s cmd =
               (tmpfileResource (view mhKeepTemp cmd))
               (\_ fp -> toProcessConfigWithTempfile (view mhMakeProcess cmd) fp) <$>
             entityToBytes e
-   in either (pure . flip setError s) (`runEntityCommand` s) mkConfig
+   in either
```
romanofski: Is it possible to simplify this? I couldn't find a solution, since `mkConfig` produces an `Either` while the `runExceptT` gives an `Either` wrapped in an `IO`.

src/Purebred/System/Process.hs
```diff
@@ -82,19 +89,44 @@ readProcess
...
+runEntityCommand cmd s =
+  let acquire = view (ccResource . rsAcquire) cmd
+      update = view (ccResource . rsUpdate) cmd
+      free = view (ccResource . rsFree) cmd
+   in do tmpfile <- acquire
+         update tmpfile (view ccEntity cmd)
+         s' <-
+           tryReadProcess (view ccProcessConfig cmd (view ccEntity cmd) tmpfile) >>=
+           (flip (view ccAfterExit cmd) tmpfile <$> handleExitCode s)
+         free tmpfile
```
romanofski: Hm... now looking at this I wonder if it's doing what I want it to do ...

src/UI/Actions.hs
frasertweedale: You might be able to simplify by having something like:
```haskell
mkConfig :: (MonadError Error m) => m (EntityCommand a FilePath)
```
And `throwError` instead of `Left`.  Then

```haskell
either handleError id $ runExceptT (mkConfig >>= runEntityCommand ec s)
```

(Something like that; just a sketch).

romanofski: Cheers! Will try it out.

src/Purebred/System/Process.hs
frasertweedale: I do wonder if we shouldn't use `bracket` (or similar) here, so that resource can be properly cleaned up if exception occur during `update` or `tryReadProcess`.

romanofski: right? I also looked at it and thought: "Hm... I don't think the `free` is called here." Haven't had a chance to verify it tho.

Will check. Cheers for the pointer. Already forgot about it.

romanofski: Used https://hackage.haskell.org/package/exceptions-0.10.2/docs/Control-Monad-Catch.html#v:bracket since the `bracket` from `Control.Exception` is `IO` only.

src/Purebred/System/Process.hs
```diff
@@ -82,19 +90,44 @@ readProcess
...
+  => EntityCommand m a
+  -> AppState
+  -> m AppState
+runEntityCommand cmd s =
+  let acquire = view (ccResource . rsAcquire) cmd
+      update = view (ccResource . rsUpdate) cmd
+      free = view (ccResource . rsFree) cmd
+   in bracket
+        (acquire >>= \tmpfile ->
+           update tmpfile (view ccEntity cmd) >> pure tmpfile)
```
frasertweedale:
```haskell
update tmpfile (view ccEntity cmd) $> tmpfile
```

romanofski: Thanks! Will change and update.

frasertweedale:
ack remains; merge when green.

PR #295, opened 2019-06-03
Feature/119 drafts in secure directory

romanofski:


.nix/hs-notmuch.nix
```diff
@@ -1,10 +1,15 @@
 { mkDerivation, base, bytestring, c2hs, deepseq, mtl, notmuch
-, profunctors, stdenv, tagged, talloc, text, time
+  , profunctors, stdenv, tagged, talloc, text, time, fetchFromGitHub
 }:
 mkDerivation {
   pname = "notmuch";
   version = "0.2.0.0";
-  sha256 = "feeafa5bdc00c1036755c92d72e7b69cb388ae9bab5064410655007cc8216ff0";
```
romanofski: This obviously needs adjusting once you merged the changes, including the setting the upper version bound in the cabal file.

src/UI/App.hs
```diff
@@ -49,11 +49,14 @@ import UI.Views
        (indexView, mailView, composeView, helpView, listOfMailsView,
         filebrowserView, focusedViewWidget, focusedViewWidgets,
         focusedViewName)
-import UI.ComposeEditor.Main (attachmentsEditor, drawHeaders)
+import UI.ComposeEditor.Main (attachmentsEditor, drawHeaders, renderConfirm)
 import Types
 
 drawUI :: AppState -> [Widget Name]
-drawUI s = [vBox (renderWidget s (focusedViewName s) <$> focusedViewWidgets s)]
+drawUI s = let ui = vBox (renderWidget s (focusedViewName s) <$> focusedViewWidgets s)
```
romanofski: Currently we're rendering all UI sequentially TV style, top to bottom. This works 99% of the time, except when we want to render a widget on top of everything. I don't know how often this will happen or if the dialog is the only instance. So this is obviously a hack at the moment :/

test/TestUserAcceptance.hs
```diff
@@ -788,30 +836,19 @@ testUserCanAbortMailComposition =
...
-            liftIO $ step "exit insert mode in vim"
-            sendKeys "Escape" (Literal "body")
+            liftIO $ step "abort mail"
+            sendKeys "q" (Literal "Keep draft?")
 
-            liftIO $ step "exit vim"
-            sendKeys ": x\r" (Regex $ "From: "
-                             <> "\"Joe Bloggs\" <joe@foo.test>")
+            liftIO $ step "choose discard"
+            --TODO: buildAnsiRegex will cause the generated Regex not
```
romanofski: This one was weird: I couldn't use the regex to match the raw output and there was nothing obviously wrong with it.

romanofski:
This is currently waiting for: https://github.com/purebred-mua/purebred-email/pull/30 to be merged and a new release is cut.

romanofski:
Alright, this should check out now. I actually plucked the patch for rendering stacked widgets and will open it in a separate pull request, since I found regressions. This is as good as is already, no need to cram more features in.

romanofski:
Hm no.. there is still something wrong with this patch. The GHC 8.6 always seems to fail consistently. I will need to dig in on what is happening there and why it's not removing the mail file. I thought the UI might show an Error, but that is not the case.

romanofski:
@frasertweedale separated out the failing regex test in https://github.com/purebred-mua/purebred/pull/295/commits/9ad7bd62534487c37b336c215fd6f2fa42424024 :)

romanofski:
Alright, I think I'm back in the game. I found the problem: The UI had been quicker repainted than the IO operation of deleting the file. I'll put a little delay in the assertion and got the test passing.

Wait with the merging of this, because I *think* I've witnessed a small regression when editing from draft mails. I'd like to check tonight if it really is one.

romanofski:
Tested to use the fix in #311  and can confirm that it solves the problem.

frasertweedale:
noice, let's merge it.

Issue #205, opened 2018-10-13
Purebred should select last e-mail in a thread if all mails are read


romanofski:
If all mails are read, Purebred currently selects the first mail in a thread. Ideally when going back to a conversation, you want to check the last recent mail.

frasertweedale:
Should be configurable via actions right.  listMoveToEnd or whatever?  Then the user can override it to get the behaviour they want.

romanofski:
Hmmm... *doh? Why haven't I thought this earlier? Will check and report back.

romanofski:
Hm... the tricky thing is that it depends on the "read" state of the mail. The key bindings currently don't have a to the list item (hm not sure if they should). At the moment the default is to pick the next unread which works well. I guess if all are read, just pick the last mail in a thread.

PR #296, opened 2019-06-08
Select last mail if all mail is read

romanofski:
When opening old threads, most of the email in the thread is read.
Most threads on mailing lists are discussions. The last mail in the
thread typically corresponds to the last state of the
discussion. Sometimes a consensus or sometimes just where we left of in
the discussion.

Purebred picked the first unread mail, which is important if there are
unread mails. However Purebred also picked the first read mail, if all
mails were read. This resulted that the user had to close the mail,
scroll down and open the last mail.

This patch selects the last mail by default if the entire thread is
read to avoid the additional key strokes to be made.

Fixes https://github.com/purebred-mua/purebred/issues/205

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -750,8 +750,9 @@ selectNextUnread =
            let
              -- find by unread tag...
              p = Notmuch.hasTag (view (asConfig . confNotmuch . nmNewTag) s)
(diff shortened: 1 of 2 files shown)


frasertweedale:
Nice, thanks!

PR #300, opened 2019-06-24
Feature/instant validation feedback

romanofski:


romanofski:
Preliminary review really. There is one test I need to fix ...

src/UI/Actions.hs
```diff
@@ -134,6 +135,33 @@ instance Scrollable 'ScrollingMailView where
 instance Scrollable 'ScrollingHelpView where
   makeViewportScroller _ = Brick.viewportScroll ScrollingHelpView
 
+-- | contexts that have an editor
```
romanofski: Wondering about this attempt, since I think it makes the rendering more explicit when it comes to looking up editors. I started off with a function which looks up the editor by `Name`, but then I need to do the whole pattern matching against other names which are not editors as well. So it's more error prone. Any ideas how this is not a good approach?

src/UI/App.hs
```diff
@@ -58,21 +55,26 @@ drawUI s = [vBox (renderWidget s (focusedViewName s) <$> focusedViewWidgets s)]
 renderWidget :: AppState -> ViewName -> Name -> Widget Name
 renderWidget s _ ListOfThreads = renderListOfThreads s
 renderWidget s ViewMail ListOfMails = vLimit (view (asConfig . confMailView . mvIndexRows) s) (renderListOfMails s)
-renderWidget s _ MailAttachmentOpenWithEditor = renderMailAttachmentOpenWithEditor s
-renderWidget s _ MailAttachmentPipeToEditor = renderMailAttachmentPipeToEditor s
+renderWidget s _ MailAttachmentOpenWithEditor =
+  renderEditorWithLabel (Proxy :: Proxy 'MailAttachmentOpenWithEditor) "Open with:" MailAttachmentOpenWithEditor s
```
romanofski: This is still a TODO to get rid of the second `Name` passed into the function. Thinking about using `HasName a` for this.

romanofski:
This passes now with the nix builds. Suppose we need to wait for another purebred-email release to make all platforms pass.

frasertweedale:
@romanofski I'll cut a purebred-email release next week.

romanofski:
No worries. I think there was still stuff I wanted to do on #295

frasertweedale:
@romanofski whoops, I don't have my Hackage passphrase with me, nor do I have access to my home machine.  So, no purebred-email release until next week.  I am going to tag the commit though :)

romanofski:
Fair enough. I could make the change on the nix side of things. If those builds pass I think it's fine to merge.

romanofski:
This one was easier to rebase, since it's the smaller change. I'll probably have to work through the other with all the various changes. Worse case I'll have to break up the change. If tests pass this one would be good to go.

romanofski:
Failure related to missing imports, which I will fix tonight. Do we still keep supporting GHC 8.2.2?

Issue #284, opened 2019-04-05
segv when opening mail


frasertweedale:
**Describe the bug**
Segv when I opened a mail.

```
Apr 05 12:26:35 T470s systemd[1]: Started Process Core Dump (PID 19268/UID 0).
Apr 05 12:26:36 T470s systemd-coredump[19269]: Process 19235 (purebred-linux-) of user 1000 dumped core.
                                               
                                               Stack trace of thread 19243:
                                               #0  0x00007fbb790aaeab raise (libc.so.6)
                                               #1  0x00007fbb790955b9 abort (libc.so.6)
                                               #2  0x00007fbb7a315931 talloc_abort.cold.19 (libtalloc.so.2)
                                               #3  0x00007fbb7a31604d _talloc_steal_loc.cold.47 (libtalloc.so.2)
                                               #4  0x00000000006a00f3 n/a (/home/ftweedal/.cache/purebred/purebred-linux-x86_64)
Apr 05 12:26:36 T470s abrtd[8533]: Size of '/var/spool/abrt' >= 5000 MB (MaxCrashReportsSize), deleting old directory 'ccpp-2019-01-22-10:24:24.235958-10186'
Apr 05 12:26:37 T470s abrt-notification[19333]: Process 19235 (purebred-linux-x86_64) crashed in _talloc_set_destructor.cold.20()
```

**To Reproduce**
Not reliably reproducible.  May be related to something begin GC'd that we need to hang on to.

romanofski:
Does it happen when you open a gazillion of threads and while purebred is calculating the length of the list, you open the first or second mail?

frasertweedale:
@romanofski good Question; worth checking that.  Did you encounter it?

romanofski:
Yeah pretty much almost reliably. Steps:
1. Search for `*`
2. Start opening the first one or two mails, b00m.

frasertweedale:
@romanofski OK awesome, that's good info!  I've got some time off this week, I'll try and sort it out :)

frasertweedale:
Info from further investigation:

- It only occurs while the background thread is traversing the spine of the thread list to determine the number of threads

- If you wait for the "num threads" update to complete (background thread is done), then execute another large search, the problem can occur (i.e. it is not restricted to the initial search)

- The problem occur upon opening the *second* unread mail (i.e. with tag `"unread"`, or whatever `nmNewTag` is).  You can precede or interleave opening of non-unread mails without triggering the abort.

[...]

PR #302, opened 2019-07-01
add support for charset plugins

frasertweedale:
Update to purebred-email >= 0.2.  Add a CharsetLookup to the
configuration, defaulting it to 'defaultCharsets' from Data.MIME.
This allows custom configurations and plugins to provide support for
additional character sets.

This requires purebred-email >= 0.2, which is not yet on hackage but is
in git: purebred-mua/purebred-email@82bc75ff878a61a21a031b155db5e380f3852788.

See also https://github.com/purebred-mua/purebred-icu, a plugin that provides
more charsets.  To try it out (cabal newstyle build), add all relevant dirs to
`cabal.project.local`, then `cabal new-install exe:purebred lib:purebred-icu`,
modify your config (see below) then run *purebred*.

Sample `cabal.project.local`:
```
packages: ., ../purebred-email, ..purebred-icu
```

Sample `~/.config/purebred/purebred.hs`:
```
import Purebred
import qualified Purebred.Plugin.ICU

main = purebred $ tweak defaultConfig
  where
  tweak =
    id  -- various config changes
    . Purebred.Plugin.ICU.enable
```

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -99,7 +99,7 @@ library
                      , filepath
                      , mtl
                      , exceptions
-                     , purebred-email >= 0.1 && < 0.2
+                     , purebred-email >= 0.2
                      , attoparsec
                      , containers
                      , mime-types
@@ -147,7 +147,7 @@ test-suite uat
   other-modules:       UAT
   default-language:    Haskell2010
   build-depends:       base
-                     , purebred-email >= 0.1 && < 0.2
+                     , purebred-email >= 0.2
                      , tasty
                      , tasty-hunit
                      , directory
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -18,7 +18,7 @@ import Data.Maybe (fromMaybe)
 import Data.List.NonEmpty (fromList)
 import System.Exit (ExitCode(..))
 
-import Data.MIME (contentTypeTextPlain, matchContentType)
+import Data.MIME (contentTypeTextPlain, defaultCharsets, matchContentType)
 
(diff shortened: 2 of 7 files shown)


romanofski:
Wow this is so cool! I'm surprised how much had to be changed, but I guess it's an integral part of an email. With that in mind it is perhaps not as surprising. Let me try it out tonight and we merge :)

romanofski:
@frasertweedale you mind if I add the nix package changes as a separate commit on top of this?

romanofski:
Yep. Works for me :)

frasertweedale:
@romanofski sure, you can add nix changes as separate commit.  We'll have to wait for purebred-email release next week before CI can go green.  Will merge after that.  **Thanks for testing!**  If I find any spare time during rest of week, I'll be focusing on the crash bug #284 .

frasertweedale:
> Wow this is so cool! I'm surprised how much had to be changed, but I guess it's an integral part of an email. With that in mind it is perhaps not as surprising. Let me try it out tonight and we merge :)

Yeah, there are so many places within MIME that charsets can appear :)

romanofski:
Ok lets wait until the release is made. Because then the nix configuration could just use those releases as well instead of pinning to certain git revisions.

frasertweedale:
OK, kicking off CI again now that purebred-email release happened.

romanofski:
Just realised updating #295 that this patch needs to go in first because it also adjusts the `Storage.ParsedMail` side of purebred, otherwise we'll end up doing double work which would be pointless.

PR #305, opened 2019-07-15
uat: extract framework to module

frasertweedale:
Lots of lovely refactoring.


frasertweedale:
@romanofski if you like it, merge it.  There are several other improvements I'd like to make and it's going to be faster to iterate if we keep it all in the purebred repo for now, and extract it only after we're both really happy with it for the purebred use case at least.

Issue #297, opened 2019-06-08
UI types should allow to express stacked widgets


romanofski:
**Is your feature request related to a problem? Please describe.**
Brick allows rendering widgets on top of other widgets. That is achieved by rendering all widgets in the background first, and then render the foreground widget on top. For example:

```Haskell
drawUI :: AppState -> [Widget Name]
drawUI s = let ui = vBox (renderWidget s (focusedViewName s) <$> focusedViewWidgets s)
           in case focusedViewWidget s of
                ConfirmDialog -> [renderConfirm s, ui]
                _ -> [ui]
```

As you can see, this "hack" above is used (from #119), because the UI has no way of expressing, that widgets need to be stacked in order to render them on top of each other.

Currently we are rendering a flat list of widgets defined in `UI.Views` and `App.hs`. Each View defines a list of widgets which are `Visible` to be rendered. The flat list allows us to conveniently flip their viewing state in `Actions.hs` whenever we want to hide or display additional widgets.

**Describe the solution you'd like**
The solution needs to use a more expressive data type which allows:
* to allow conveniently manipulate the viewing state, while
* express the stacking order of the widgets

**Describe alternatives you've considered**
Don't know how exactly would work and which data types would be suitable.

**Additional context**
None

romanofski:
I thought about this a bit. Currently we have:
```Haskell
data View = View
    { _vFocus :: Name
    , _vWidgets :: Tiles
    }

-- A tile is a view element with a widget and it's view state
newtype Tiles =
  Tiles (V.Vector Tile)
  deriving (Eq, Show)

type instance Index Tiles = Name
type instance IxValue Tiles = Tile

instance Ixed Tiles where
  ix = tile
```
and I thought, well we could express the stacked nature by making the Tiles a list of lists like:

```Haskell
newtype Tiles =
  Tiles (V.Vector (V.Vector Tile))
  deriving (Eq, Show)
```

(perhaps with a better data type). The view definition would then become from:

```Haskell
composeView :: View
composeView =
  View
    { _vWidgets =
        Tiles $ fromList
          [ Tile Visible ComposeHeaders
          , Tile Visible ComposeListOfAttachments
          , Tile Visible StatusBar
          , Tile Hidden ComposeFrom
          , Tile Hidden ComposeTo
          , Tile Hidden ComposeSubject
          , Tile Hidden ConfirmDialog
          ]
    , _vFocus = ComposeFrom
    }
```
to
```Haskell
composeView :: View
[...]

PR #312, opened 2019-07-28
Allow stacked rendering of widgets

romanofski:
Brick supports rendering of stacked widgets. For example, Brick can
render a dialog in front of the main UI.

Before this patch, Purebred rendered widgets in a single list onto the
screen. The widgets were rendered vertically from top to
bottom. However, if we wanted to render a dialog in front of the UI it
was rendered at the top and the rest of the widgets used the remaining
height showing them a little bit squished.

This patch changes this by adding an additional abstraction. Instead of
a single list of widgets, purebred allows lists of widgets. Layers can
not be set to hidden or visible, but the containing widgets can.

Each layer renders it's visible widgets resulting in a stacked, rendered
terminal image.

Fixes https://github.com/purebred-mua/purebred/issues/297

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -452,33 +452,36 @@ veName f (Tile a b) = fmap (\b' -> Tile a b') (f b)
 veState :: Lens' Tile ViewState
 veState f (Tile a b) = fmap (\a' -> Tile a' b) (f a)
 
+type Layers = V.Vector Layer
+
 data View = View
     { _vFocus :: Name
-    , _vWidgets :: Tiles
+    , _vLayers :: Layers
     }
 
--- A tile is a view element with a widget and it's view state
-newtype Tiles =
-  Tiles (V.Vector Tile)
-  deriving (Eq, Show)
-
-type instance Index Tiles = Name
-type instance IxValue Tiles = Tile
-
-instance Ixed Tiles where
-  ix = tile
-
-tileiso :: Iso' Tiles (V.Vector Tile)
-tileiso = iso (\(Tiles xs) -> xs) Tiles
-
-tile :: Name -> Traversal' Tiles Tile
-tile k = tileiso . traversed . filtered (\x -> k == view veName x)
-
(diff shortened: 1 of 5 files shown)


src/Types.hs
```diff
@@ -452,33 +452,36 @@ veName f (Tile a b) = fmap (\b' -> Tile a b') (f b)
...
 
-tile :: Name -> Traversal' Tiles Tile
-tile k = tileiso . traversed . filtered (\x -> k == view veName x)
+instance Ixed Layer where
+  ix = tile
 
-vWidgets :: Lens' View Tiles
-vWidgets = lens _vWidgets (\settings x -> settings { _vWidgets = x })
+layeriso :: Iso' Layer (V.Vector Tile)
+layeriso = iso (\(Layer xs) -> xs) Layer
```
frasertweedale: Don't bother changing the patch, but I just want to draw your attention to http://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Iso.html#v:coerced which is an automatic iso between newtypes and the underlying type.

romanofski: I have trouble with the typing tho. For example:

```Haskell
tile :: Name -> Traversal' Layer Tile
tile k = layeriso . traversed . filtered (\x -> k == view veName x)
```

If I replace `layeriso` with `coerced` I get:

```
src/Types.hs:484:10-16: error:
    • Couldn't match representation of type ‘V.Vector Tile’
                               with that of ‘f0 Tile’
        arising from a use of ‘coerced’
    • In the first argument of ‘(.)’, namely ‘coerced’
      In the expression:
        coerced . traversed . filtered (\ x -> k == view veName x)
      In an equation for ‘tile’:
          tile k = coerced . traversed . filtered (\ x -> k == view veName x)
    |
484 | tile k = coerced . traversed . filtered (\x -> k == view veName x)
    |          ^^^^^^^
```

If I use coerced for the `layeriso` implementation everything checks out. What's the better way to apply it in this situation?

PR #317, opened 2019-08-02
Fix/nix include purebred icu

romanofski:


frasertweedale:
I mean... I dunno if forcing use of *purebred-icu* is a solution.  We can't do it for every conceivable plugin...

But I don't mind if you merge this, since I don't use nix yet :)

romanofski:
Hm.. I actually don't know why travis is not running any tests o.O

In regards to forcing the plugin... I guess ideally we should perhaps be using multiple configs instead of a single one. That way there is no need of always enabling it. Alternatively move some of this testing into purebred-icu since we have the UAT framework in a separate package.

romanofski:
Let's sit on this one for a bit. I might just keep the nix changes, but not enable it globally. Because you're right: can't do it for all.

romanofski:
I removed the changes to our tests and CI. Made the purebred-icu integration optional, which can be enabled with an argument to `nix-shell` or `nix-build`.

Issue #182, opened 2018-06-29
Ways to edit attachments


romanofski:
We need a way to view/edit attachments.

My initial thoughts were to add a parser for a mailcap file, but in 2018 perhaps there are better ways.

romanofski:
After implementing #10, I think what would be useful to add in the configuration is a:

```Haskell
Map ContentType (ProcessConfig ()()())
```
Meaning, if we can lookup the ProcessConfig type (which supports `IsString`) by the given `ContentType` use it as a command, otherwise prompt the user.

Perhaps a Plugin could be later utilized which could for example parse an existing `mailcap` file and provide the `Map` ...

romanofski:
I looked into this. The tricky bit could be `ProcessConfig` and using the above type, because we're getting into trouble with `NFData`:
```
src/Types.hs:348:24-29: error:
    • No instance for (NFData (ProcessConfig () () ()))
        arising from the 'deriving' clause of a data type declaration
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (NFData MailViewSettings)
    |
348 |     deriving (Generic, NFData)
    |                        ^^^^^^
```
It seems `ProcessConfig` is only exported as a type from the API which will make it harder for us to declare the `NFData` instance.

frasertweedale:
@romanofski I think we want a richer type anyway - one that we can easily convert into a `ProcessConfig` - that allows more fine-grained control over where the filename appears, and supports pipes (perhaps).  For now I would say make the simplest data type that expresses what we currently support, create an `NFData` instance for it, and put that in the config.  Then we convert that to a `ProcessConfig` just in time to actually use it.

Alternatively, just make the `NFData` instance of the config record type that immediately includes the `ProcessConfig` values skip the `ProcessConfig` values.  It's not lawful, and not ideal, but there is precedent: see `instance NFData (Keybinding v ctx)` in `Types.hs` (there is no `NFData` instance for `Vty.Event` so I skip it.)

romanofski:
I tried the latter, but that means I can pattern match on the `ProcessConfig` constructor, yes? Since the constructor is not exported I think I need to pick the first idea.

frasertweedale:
@romanofski write the `NFData` instance for the thing that contains the `ProcessConfig`, and just don't `force` the `ProcessConfig` value.  There's no need to pattern match.

Push your WIP and let me help if you are unsure how to continue.

[...]

PR #318, opened 2019-08-03
WIP Fix/182 edit attachments

romanofski:


romanofski:
@frasertweedale I'd be very glad if you could give me some pointers :)

Obviously don't merge since it's my current hack to get a better understanding how `HasList` works. I'm currently a bit stuck since I can't make sense on how to make the compiler happy and I also don't know what to google for in order to read up on it. 

I want to use `HasList` in order to look up the attachments from two different locations. One gives me a `WireEntity` and another gives me a `MIMEMessage`. In order to solve this, I use a typeclass `Foo a` (both instances currently not implemented, just to see if it's checks out) and pass the converted `ByteString` to the `EntityCommand`. What I get is:

```
src/UI/Actions.hs:1218:66-77: error:
    • Could not deduce (Foo (E ctx))
        arising from a use of ‘toByteString’
      from the context: (HasList ctx, Foldable (T ctx),
                         L.Splittable (T ctx), MonadError Error m, MonadIO m)                                                         
        bound by the type signature for:
                   selectedAttachmentOrError :: forall (ctx :: Name) (m :: * -> *).                                                   
                                                (HasList ctx, Foldable (T ctx),                                                       
                                                 L.Splittable (T ctx), MonadError Error m,                                            
                                                 MonadIO m) =>
                                                Proxy ctx -> AppState -> m C8.ByteString                                              
        at src/UI/Actions.hs:(1204,1)-(1213,19)
    • In the first argument of ‘to’, namely ‘toByteString’
      In the second argument of ‘(.)’, namely ‘to toByteString’
      In the second argument of ‘(.)’, namely ‘_2 . to toByteString’                                                                  
     |
1218 |         (list proxy . to L.listSelectedElement . _Just . _2 . to toByteString)                                                 
     | 
```
It is a bit puzzling to me since we explicitly state what `E ctx` is in each instance. So, what is missing here, that the compiler can't deduce it?

frasertweedale:
@romanofski a bit much for me to wrap my head around.  Maybe we should get together for a hack day soon?

romanofski:
Aaah yes let's do that.

romanofski:
I went back to this and I think what I'm trying to do doesn't work. I can't explain why tho. I was trying to convert the result of a lens, e.g.
```Haskell
view (asMailView . mvAttachments . to L.listSelectedElement . _Just . _2) s
```
but abstracted as
```Haskell
view (list (Proxy :: Proxy ctx) . to L.listSelectedElement . _Just . _2)
```
which would give me a `MIMEMessage` for instance, but instead I get the type family synonym `E ctx`. The goal would have been to be able to convert both `MIMEMessage` and `WireEntity` to a `ByteString` for writing them both to the file system in order to open or edit them.

However, thinking about it now I wonder if the whole approach is flawed. Specifically the fact that I'm dealing one with almost a `ByteString` (the `WireEntity` who's body is a `ByteString`) and a possibly encapsulated mime tree (the `MIMEMessage`). Converting a `WireEntity` seems easy enough, but converting a `MIMEMessage` to a `ByteString` for easy editing seems like trouble. For mail composition I could imply that we're only dealing with one attachment, but what happens if we're forwarding a mail. The whole assumption breaks down.

I presume I'll have to think about this a bit more where to draw the line here. Should we maybe move the attachments when composing mails to use `WireEntity`s? Does that maybe just shift the problem a bit down the road (e.g. what happens when we forward mails ...).

romanofski:
@frasertweedale I thought about this more and I wonder what you think. I think I'm over-engineering this quite a bit where there is a simpler solution. 

At the moment we store attachments of the parsed mail as `WireEntity` and "convert" it into the right format upon saving/opening/viewing whatever. Composition is different in such a way, that we compose a `MIMEMessage` with attachments with multiple `MIMEMessage`s with a single `Part`.

I'm proposing to change the latter and also move to using a `WireEntity` when composing and attaching files.

Apart from the fact that would obviously make the code less complex, we could re-use the code for opening/saving already used for parsed messages. I have not measured, but I think it would provide an additional performance gain of simply keeping the body of an attachment as a `ByteString` until we actually send it. But the biggest argument for the change I think is the fact that the current form used for attachments, `MIMEMessage`, is too arbitrary to know how many attachments it actually holds. Theoretically it could hold a tree or many additional `Part`s, but implicitly I would have to trust that we never screw up the code to only hold one part (which it does now) - the attachment itself.

Does that reasoning make sense?

PS: I think it probably makes sense to keep some of the functions, eg. `createTextPlainMessage` to prepare a quick message. It would just not be useful for purebred.

frasertweedale:
> I'm proposing to change the latter and also move to using a `WireEntity` when composing and attaching files.
> 
> Apart from the fact that would obviously make the code less complex, we could re-use the code for opening/saving already used for parsed messages. I have not measured, but I think it would provide an additional performance gain of simply keeping the body of an attachment as a `ByteString` until we actually send it. But the biggest argument for the change I think is the fact that the current form used for attachments, `MIMEMessage`, is too arbitrary to know how many attachments it actually holds. Theoretically it could hold a tree or many additional `Part`s, but implicitly I would have to trust that we never screw up the code to only hold one part (which it does now) - the attachment itself.
> 
> Does that reasoning make sense?
> 
> PS: I think it probably makes sense to keep some of the functions, eg. `createTextPlainMessage` to prepare a quick message. It would just not be useful for purebred.

I see two main approaches.

### Approach A

(This seems to line up with your proposal).

During the compose phase, we store (in the program state) the message being composed with a different type, e.g.:

```haskell
type ComposeMessage = Message () (NonEmpty WireEntity)
```

This enforces a "flat" structure.  Conversion to `MIMEMessage` is trivial.  It can be defined in *purebred* (doesn't have to go into *purebred-email*).

Of course, any of the `WireEntity` values could be a serialised `MIMEMessage` (a nested multipart message or encapsulated message).  As you rightly point out, editing that as a ByteString is dangerous.  There are two ways to deal with this:

1. Don't deal with it; treat it as a caveat, move on.
2. Enhance the "mailcap" editing interface to support *validation* of the result of an edit, such that invalid edits can be rejected.  This is of limited 

Of course, we can start with **1** and implement **2** later.

### Approach B

Continue using `MIMEMessage` but present it as a tree structure, and only allow leaf entities to be edited.  This needs some sort of tree view widget for Brick, either generic (implying some sort of "tree view" type class to be designed, and instantiated for `MIMEMessage`) or specialised to `MIMEMessage`.

romanofski:
@frasertweedale ah cheers for answering. It actually highlighted an aspect I didn't think about. It sounds Approach (B) would be the better choice. We know that we need a tree view anyway, so it's more or less a matter when we implement it. Being able to edit leaf entities with a tree widget would actually be really spiffy too.

Which brings us what we do with closing off #182 . We could:
a) close it and create a specific card of editing attachments during composition using the tree widget and basically implement once we have a tree widget
b) we really think it's a 1.0 feature (not so sure really, just remove and re-attach) and implement it keeping `MIMEMessage` and only pick the first `Part`. Obviously acknowledging that it's not a complete feature.

Personally I think postponing and getting the tree widget first is my favorite (a). Do you agree?

frasertweedale:
@romanofski yes I agree.

Issue #264, opened 2019-03-25
Clean up rendering of WireEntity and MIMEMessage (DRY)


romanofski:
**Is your feature request related to a problem? Please describe.**
With merging of #257 I introduced a duplicated function `renderPart` to render the `WireEntity` in `UI.Mail.Main`. Almost the same code is used in `UI.ComposeEditor.Main` with the same name.

We should remove the duplication and use one function.

**Describe the solution you'd like**
Don't repeat yourself but have one function which can render a `WireEntity` and a `MIMEMessage` in the UI.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #320, opened 2019-08-04
DRY: use one type class to get MIME information

romanofski:
We needed two separate functions in order to gather information from
MIMEMessage or WireEntity in order to render them as a list of
attachments. Both are used in separate views: the former when viewing
mails stored in notmuch the latter when constructing new mails.

This patch adds a unified interface so we can use one function to render
both.

Fixes: https://github.com/purebred-mua/purebred/issues/264

### DIFF
--- a/src/UI/Mail/Main.hs
+++ b/src/UI/Mail/Main.hs
@@ -3,6 +3,7 @@
 module UI.Mail.Main
   ( renderMailView
   , renderAttachmentsList
+  , renderPart
   ) where
 
 import Brick.Types (Padding(..), ViewportType(..), Widget)
@@ -14,7 +15,6 @@ import Brick.Widgets.Core
 import Control.Lens (filtered, folded, toListOf, view, preview, has)
 import qualified Data.ByteString as B
 import qualified Data.CaseInsensitive as CI
-import Data.Maybe (fromMaybe)
 
 import Data.MIME
 import Storage.ParsedMail (chooseEntity, entityToText)
@@ -22,6 +22,7 @@ import Storage.ParsedMail (chooseEntity, entityToText)
 import Types
 import UI.Draw.Main (attachmentsHeader)
 import UI.Views (focusedViewWidget)
+import UI.Utils (takeFileName)
 import Config.Main (headerKeyAttr, headerValueAttr, mailViewAttr,
                     listSelectedAttr, listAttr)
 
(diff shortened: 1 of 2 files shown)


romanofski:
Huh... randomly failing tests? I think I messed up showing only the filename which causes the random failures. Some paths are long enough to match the substring, some are not and then we have a failure ...

src/Purebred/Mail.hs
```diff
@@ -0,0 +1,55 @@
...
+
+import UI.Utils (takeFileName)
+import Data.MIME
+
+-- | Provide metadata used for render the message using a unified
+-- interface
+class MIMEData a where
+  contentType :: a -> T.Text
+  filename :: CharsetLookup -> a -> T.Text
+  attachmentType :: a -> T.Text
```
frasertweedale: I'm really confused as to why this abstraction is necessary.  All of these only read from the headers, right?
`MIMEMessage` and `WireEntity` are both type synonyms for `Message something something'`, so you should be able to use *exactly* the same functions to read these data from either type.

romanofski: hrm.. odd.. I'll have another look. Cheers!

src/UI/Mail/Main.hs
```diff
@@ -67,25 +68,22 @@ renderAttachmentsList :: AppState -> Widget Name
...
       widget = hBox
         [ padLeftRight 1 attachmentType
         , padRight Max (txt pFilename)
         , txt pType
         ]
   in withAttr listItemAttr widget
 
-isAttachment' :: WireEntity -> Bool
-isAttachment' = has (headers . contentDisposition . dispositionType . filtered (== Attachment))
+isAttachment' :: Headers -> Bool
```
romanofski: This is currently implemented in purebred-email as:
```Haskell
isAttachment :: MIMEMessage -> Bool
```
but I think we should change it to accept `Headers` as the first argument. Do you concur?

frasertweedale: In my 'stuff' branch it is already generalised to `isAttachment :: HasHeaders a => a -> Bool`.

`Headers` and `Message s a` have instances of `HasHeaders`.

So... soon :)

romanofski:
Oh @frasertweedale I think this is really good to go. I'll merge if you don't mind.

frasertweedale:
@romanofski go ahead.

Issue #286, opened 2019-04-21
Notification system


romanofski:
**Is your feature request related to a problem? Please describe.**
Showing errors, warnings or informational messages in the UI is currently a simple stub. Everything is displayed as an error if a message for it is set in the `AppState`. We need something more flexible and more nuanced.

**Describe the solution you'd like**
I'd like to have a way:
1. to distinguish the severity of a message, e.g. warning, notification, error
2. to display them
3. to remove them

I think 2) and 3) are the most ambiguous at this point. At the moment we always clear the error message at every key press. I wonder if it's not better for more severe notifications to have a timer. Having said that, the problem with a timer is that it becomes more complex that the current state, since what happens if the user presses the action leading to the error more than once. Prolong the delay of remove the message?

Perhaps to keep things simple, we keep the removal of the message on every key press as it is now, and cross that bridge of a more complex system when we need it.

**Describe alternatives you've considered**
It's vague as it is.

**Additional context**
None

romanofski:
I just noticed that I made a mistake with #323 mistakenly closing this one. This still needs to be done.

PR #323, opened 2019-08-26
New mail indicator

romanofski:
This patch adds a small indicator which shows how much new mail is in
the INBOX. New mail is defined by a customizable notmuch query. The
query is rerun after a configurable delay. By default it is 5 seconds.

Fixes https://github.com/purebred-mua/purebred/issues/286

### DIFF
--- a/src/Config/Main.hs
+++ b/src/Config/Main.hs
@@ -193,6 +193,8 @@ defaultConfig =
       , _nmNewTag = "unread"
       , _nmDraftTag = "draft"
       , _nmSentTag = "sent"
+      , _nmHasNewMailSearch = "tag:inbox and tag:unread"
+      , _nmHasNewMailCheckDelay = Just (5 * 10000 * 60)  -- 3 seconds
       }
     , _confEditor = fromMaybe "vi" <$> lookupEnv "EDITOR"
     , _confMailView = MailViewSettings
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -189,6 +189,12 @@ getDatabasePath = do
   where
       decode = T.unpack . sanitiseText . decodeLenient . LB.toStrict
 
+countThreads ::
+     (MonadError Error m, MonadIO m) => T.Text -> FilePath -> m Int
+countThreads query fp =
+  withDatabaseReadOnly fp $
+  flip Notmuch.query (Notmuch.Bare $ T.unpack query)
+  >=> Notmuch.queryCountMessages
 -- | creates a vector of threads from a notmuch search
 --
 getThreads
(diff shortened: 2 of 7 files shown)


romanofski:
Not really sure where to draw the line here. So this one runs every 5 seconds ... I first thought of running it every 5 minutes, but then the question is what keeps it updated once you start archiving some of the unread mails.

new

This patch doesn't have any tests yet... so they still need to be added. I also haven't looked through the code yet...

src/Config/Main.hs
```diff
@@ -195,6 +195,8 @@ defaultConfig =
       , _nmNewTag = "unread"
       , _nmDraftTag = "draft"
       , _nmSentTag = "sent"
+      , _nmHasNewMailSearch = "tag:inbox and tag:unread"
+      , _nmHasNewMailCheckDelay = Just (5 * 10000 * 60) -- ^ 5 seconds
```
romanofski: Thought about this and I think I'll better use a custom data type like: 
```Haskell
data Delay = Seconds Int
                       | Minutes Int
```
I can see myself forgetting a zero or two and then you have a thread polling notmuch like crazy.

romanofski: Good idea actually. I'll keep it unresolved until I've updated the patch with some tests.

src/Storage/Notmuch.hs
```diff
@@ -189,6 +189,12 @@ getDatabasePath = do
   where
       decode = T.unpack . sanitiseText . decodeLenient . LB.toStrict
 
+countThreads ::
+     (MonadError Error m, MonadIO m) => T.Text -> FilePath -> m Int
```
romanofski: I'm wondering from a "library" perspective if I shouldn't pass in the configuration, since most other functions in this module expect it. Is it better for consistency sake?

src/UI/Status/Main.hs
```diff
@@ -33,6 +38,20 @@ import UI.Views (focusedViewWidget, focusedViewName)
...
 import Error
 import Config.Main (statusbarAttr, statusbarErrorAttr)
+import qualified Storage.Notmuch as Notmuch
+
+checkForNewMail :: BChan PurebredEvent -> FilePath -> Text -> Int -> IO ()
+checkForNewMail chan dbpath query delay = do
+  r <- runExceptT (Notmuch.countThreads query dbpath)
+  case r of
+    Left _ -> pure ()
+    Right n -> notify n >> rescheduleMailcheck chan dbpath query delay
```
frasertweedale: Please use `*>`.

src/UI/Status/Main.hs
```diff
@@ -33,6 +38,20 @@ import UI.Views (focusedViewWidget, focusedViewName)
...
+  where
+    notify = writeBChan chan . NotifyNewMailArrived
+
+rescheduleMailcheck :: BChan PurebredEvent -> FilePath -> Text -> Int -> IO ()
+rescheduleMailcheck chan dbpath query delay =
+  void $ forkIO (threadDelay delay >> checkForNewMail chan dbpath query delay)
```
frasertweedale: and here :)

PR #325, opened 2019-09-14
bump purebred-email to v0.3 and chase API changes

frasertweedale:


romanofski:
Btw. push when all the other builds except the nix build fails. I'll fix the nix part up.

romanofski:
Looks like some acceptance tests are failing because we're displaying a bit more information. I'd be happy to fix them up and the nix part too if you don't mind so we can merge this :)

frasertweedale:
@romanofski feel free; I might get to it this weekend or maybe not, but if you have time go for it.

PR #329, opened 2019-09-17
Remove additional theme

romanofski:
I personally don't use it. Maintaining is additional work I'm currently
not willing to spend. It isn't really a game changer if we don't ship
two themes.


frasertweedale:
Agree.  :ship:

Issue #69, opened 2017-09-26
add license blurb to files


frasertweedale:
source files should have a license blurb in them, 

```
-- This file is part of purebred
-- Copyright (C) 2017  {{Fraser Tweedale and/or Roman Joost}}
--
-- purebred is free software: you can redistribute it and/or modify
-- it under the terms of the GNU Affero General Public License as published by
-- the Free Software Foundation, either version 3 of the License, or
-- (at your option) any later version.
--
-- This program is distributed in the hope that it will be useful,
-- but WITHOUT ANY WARRANTY; without even the implied warranty of
-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-- GNU Affero General Public License for more details.
--
-- You should have received a copy of the GNU Affero General Public License
-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

I'll get around to this; just making a note so I don't forget :)

romanofski:
Yeah good idea!

romanofski:
Btw. just realising in regards to the author. I'd be in favor of Fraser and Roman for the authorship ... In the future it will go hopefully: and contributors ?

frasertweedale:
Yeah just add author names to the copyright statement in each file where a substantive
contribution was made by each author.

romanofski:
Just realising that github shows the author of the file, as well as it's contributors. Perhaps it's something to make use of?

frasertweedale:
Well, so does git-blame.  We can automate it, but it's probably quicker to do it by hand :)

PR #330, opened 2019-09-17
Add license blurb to remaining files

romanofski:
Fixes: https://github.com/purebred-mua/purebred/issues/69


romanofski:
Was looking for something quick assuming only a few files were left. One mistake led to another ...

frasertweedale:
Thanks!

Issue #333, opened 2019-10-21
Cursor position on compose editor fields isn't at the end


romanofski:
**Describe the bug**
When you reply to an e-mail and want to change e.g. the `From` field, the cursor position is at the beginning and not at the end.

**To Reproduce**
1. Go to an e-mail of your choice
2. Hit reply (`r`)
3. Fill out the mail and exit the editor
4. Edit the `From` field

**Expected behavior**
The position of the cursor should be at the end of the editor.

**Additional context**
Applies to every editor here. The `Subject` and `To` fields are also affected.

PR #337, opened 2019-10-24
Fix/331 editors arent hidden

romanofski:


romanofski:
Includes the fix for #333 too.

PR #339, opened 2019-11-05
add CLI option to override initial search

frasertweedale:
Sometimes you want to start with a different search from what is in
your user config (without changing the user config, triggering
recompile).  Add the --search CLI option to override the initial
search term.


romanofski:
Nice!

PR #344, opened 2019-11-21
Also display the confirm dialog when using Esc

romanofski:
This was a forgotten binding. When the user presses ESC he looses his
entire Draft instantly.

Avoid this by binding the same Action to the two different keys.


frasertweedale:
Should have UAT for this?

romanofski:
I was just about to say that we don't need one since we already have a few using `q`. I'll changed one of the UATs to use `Escape` instead. Result should be the same.

PR #346, opened 2019-11-22
Fix/327 intermittend test failure

romanofski:


romanofski:
@frasertweedale gave me a semi-go ahead on IRC... so I go ahead :)

PR #347, opened 2019-11-26
Piped output (Autoview)

romanofski:
This adds functionality to pipe output through an external program and
capture it's output for display. This allows searching and/or viewing
mails with non-text entities like... HTML.

While it has "Autoview" in the name, this patch isn't delivering what
mutt or any other mail provides. Namely showing the content of all
attachments if they can be piped through an external program. For this,
there are several outstanding things to be implemented like:

a) Extend the AST to actually represent attachments
b) Try to traverse these attachments in order to produce text output
c) Extend the search functionality, most specifically the `ScrollSteps`
   to allow searching through the AST body including the output of the
   attachments.

The entire reason why this is put up for a pull request is the fact that
it already provides some relieve to show non-text mails, especially in
an environment using heavily HTML mails for example.

### DIFF
--- a/src/Storage/ParsedMail.hs
+++ b/src/Storage/ParsedMail.hs
@@ -15,6 +15,7 @@
 -- along with this program.  If not, see <http://www.gnu.org/licenses/>.
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections #-}
 
 module Storage.ParsedMail (
   -- * Synopsis
@@ -47,6 +48,8 @@ import Control.Lens
 import Data.Text.Lens (packed)
 import Control.Monad.Except (MonadError, throwError)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import Control.Monad.Catch (MonadMask)
+import Data.Foldable (toList)
 import qualified Data.ByteString as B
 import qualified Data.CaseInsensitive as CI
 import qualified Data.Text as T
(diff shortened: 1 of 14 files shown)


src/Storage/ParsedMail.hs
```diff
@@ -166,38 +198,58 @@ entityToText charsets msg = sanitiseText . either err (view body) $
...
+    (tmpfileResource (view mhKeepTemp cmd))
+    (\_ fp -> toProcessConfigWithTempfile (view mhMakeProcess cmd) fp)
+    tryReadProcessStdout
+
 quoteText :: T.Text -> T.Text
-quoteText = T.unlines . fmap ("> " <>) . T.lines
+quoteText = ("> " <>)
 
+-- | Creates a new instance of `MIMEMessage` with a quoted plain text part if:
+-- a) the preferred content type can be extracted
```
romanofski: Ugh.. what happened here. Will need to fix this up.

frasertweedale:
@romanofski conflicts

romanofski:
eek! Should be good again ...

src/Types.hs
```diff
@@ -784,14 +798,25 @@ data MakeProcess
             [String]
   deriving (Generic, NFData)
 
+mpCommand :: Lens' MakeProcess (NonEmpty Char)
+mpCommand f (Shell x) = fmap (\x' -> Shell x') (f x)
+mpCommand f (Process x args) = fmap (\x' -> Process x' args) (f x)
+
 data MailcapHandler = MailcapHandler
   { _mhMakeProcess :: MakeProcess
+  , _mhCopiousoutput :: Bool
```
frasertweedale: OK, I think we need to define new data types instead of using Bool here.  One Bool in a product type is maybe okay.  But when your constructions start looking like `FooThing "schwa" True False False` it is hard to know what the True and False are about, and easy to make mistakes.

So I propose new data types:
```haskell
data CopiousOutput = CopiousOutput | NoCopiousOutput
data KeepTempOnExit = KeepTempOnExit | NoKeepTempOnExit -- or DiscardOnExit, or whatever
```

frasertweedale:
OK with me (not that I actually tried it...).  One nit, which I'm fine to either address here or file an easyfix issue for it.

src/Types.hs
romanofski: That is true. Thanks for the review. Much appreciated!

PR #348, opened 2019-11-26
Fix regression of packages

romanofski:
The b698d93f22e7f7f469b7ed25f2c1df9319df4868 commit regressed our
packages in that: it still built purebred, but it had to be split in
such a manner that when a user installs 'purebred', the development
packages come with it. This is important, since the configuration needs
to be recompiled with 'Purebred'.

This patch fixes this problem, but re-introducing the subpackage
structure again.

### DIFF
--- a/purebred.spec
+++ b/purebred.spec
@@ -8,7 +8,7 @@
 
 Name:           %{pkg_name}
 Version:        0.1.0.0
-Release:        0.20191104%{?dist}
+Release:        0.20191125%{?dist}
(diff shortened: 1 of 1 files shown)


romanofski:
The last update just failed with the problem it supposed to fix. Maybe I need to go back and think this over...

romanofski:
> The last update just failed with the problem it supposed to fix. Maybe I need to go back and think this over...

Ignore this. I thought this patch was addressing the intermittend failure ...

Issue #352, opened 2019-12-11
Generic build instructions


hughwilliams94:
Would it be possible to provide some straightforward build instructions for user that are not using Fedora or the Nix package manager. I tried to look through Travis config and run the cabal commands manually, but always encountered problems on Arch Linux.

romanofski:
Hi @hughwilliams94 ,

I think that's a great idea. I think our HACKING document provides insights in how to setup a development environment, but is perhaps not as obvious as our README on the front of the repository?

frasertweedale:
On Thu, Dec 12, 2019 at 04:24:54PM -0800, Roman Joost wrote:
> Hi @hughwilliams94 ,
> 
> I think that's a great idea. I think our HACKING document provides insights in how to setup a development environment, but is perhaps not as obvious as our README on the front of the repository?
> 
Maybe a link to HACKING from README will suffice (plus whatever
updates we might need to make to HACKING)?

hughwilliams94:
I did have a look at the HACKING document, but again the instructions are for using Nix. I do agree it should probably be linked from the readme though.

romanofski:
The thing with linking it from the README is that I think no one actually notices it. Reason that I believe that it's not obvious is because we already link to the HACKING document.

Perhaps the better place (I thought at least) would be below the repositories/instructions for Fedora/Nix ?

romanofski:
@hughwilliams94 Btw. if there are any specific problems you're hitting when you've tried to build purebred I'd be interested to know about them too. Can't hurt to know what people try and what they're running into especially on platforms we don't use.

romanofski:
@hughwilliams94 would you mind giving the instructions added in #353 a crack? I'd be interested whether these are enough to help you get going.

hughwilliams94:
Hi thanks for updating the page. I had a go but I have been getting endless dependency errors from `cabal`. I think it may well be linked to this point in the [Arch Wiki](https://wiki.archlinux.org/index.php/haskell#Problems_with_linking). I don't have any idea how the haskell packaging ecosystem works, but I tried the recommendations in that link and nothing seems to help.

hughwilliams94:
Reading up a little it seems that Haskell is a bit messed up on Arch, I have decided to remove all of the system packages and try to go down the `stack` route to see whether the situation improves.

hughwilliams94:
[...]

PR #353, opened 2019-12-20
Add generic build instructions (Cabal)

romanofski:
This patch adds build instructions to use with cabal.

This is mainly targeting users who don't have access to nix or use a
Fedora distro. Cabal is somewhat universal to Haskell so theoretically
using Cabal should work everywhere. However, these instructions don't
give any help in regards problems users could face like:

* missing system dependencies
* GHC problems
* version mismatches
* etc

Fixes https://github.com/purebred-mua/purebred/issues/352

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ and `nix-env` to install the application:
 $ nix-env --file default.nix --install
 ```
 
+### Arch Linux
+
+We officially don't support Arch Linux, but will support any packaging
+efforts.  Please see
+[#352](https://github.com/purebred-mua/purebred/issues/352#issuecomment-567873060)
+for links to help building Purebred.
+
 ### Cabal
 
 Install development packages for system library dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,3 +69,33 @@ and `nix-env` to install the application:
...
+
+* libtalloc-devel
+* notmuch-devel
+
+Make sure you have at least one of our supported GHC and Cabal-install >= 2.2
+installed. From a cloned checkout you then go:
+
+```
+# Updates the package list needed for cabal to download dependencies
+$ cabal update
```
frasertweedale: ``should be new-update``, to go with ``new-build``

romanofski: See.. I thought that was one of the commands which didn't change. Will update.

hughwilliams94:
As I mentioned in #352 it might be worth adding a link to that issue to help any future arch users looking to build the project.

romanofski:
I've added an additional patch linking to the issue with the Arch Linux workaround. Hopefully it'll help more people to get started until someone picks up maintainer ship (and I guess we've released our first release).

Issue #313, opened 2019-07-28
Save attachments


romanofski:
**Is your feature request related to a problem? Please describe.**
With implementing #10 and #118 I forgot to add the feature of saving attachments. Since both issues were quite broad, I'll refrain from re-opening them and just file this as a new feature.

**Describe the solution you'd like**
Ideally I'd like to use the file browser to pick the folder on where to save the attachment. Perhaps to keep this small for 1.0 an editor entry is fine? The editor however is currently not able to autocomplete folders and files so just entering whatever with error messages is kind of meh too. Maybe we could investigate if solving #280 using haskeline would give us a much better editor-only solution.

**Describe alternatives you've considered**
None

**Additional context**
None

romanofski:
Lets keep this simple and only use a normal editor for entering the filepath. Perhaps we can pre-fill the path using the systems tempdir and the file name from the attachment.

PR #359, opened 2020-01-20
Allows to save attachments

romanofski:
This implements the ability to save mail attachments to disk. The editor
isn't very fancy. This is the bare minimum to be able to save any type
of attachment. Errors are simply propagated.

Fixes: https://github.com/purebred-mua/purebred/issues/313

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -77,6 +77,7 @@ module UI.Actions (
   , createAttachments
   , openAttachment
   , setTags
+  , saveAttachmentToPath
 
   -- ** Actions for scrolling
   , scrollUp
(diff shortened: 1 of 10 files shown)


src/UI/Actions.hs
```diff
@@ -759,6 +789,24 @@ pipeToCommand =
     in Brick.suspendAndResume $ liftIO $ pipeCommand' s cmd
   }
 
+-- | Save a currently selected attachment to the given path on
+-- disk. Will error if the path is invalid or can not be written
+-- to. Otherwise, write out the attachment to a temporary file and then
+-- rename it to the path given by the user.
```
frasertweedale: The doc says you write to temp file then rename.  But this doesn't look like what is happening in `writeEntityToPath`.  Did I miss something, or is the doc wrong?

romanofski: Oof why don't I see those things.

I had it implemented first with that in mind, but then decided that it was overkill and got rid of it. I'll adjust the doc string.

romanofski:
I took the liberty to merge it, having fixed up the wrong haddock description. Thanks for spotting it :)

Issue #294, opened 2019-05-30
Conditional Actions


romanofski:
**Is your feature request related to a problem? Please describe.**
I would like to chain actions which perform independent actions based on a condition. For example: `openAttachment :: Action 'ViewMail ctx (T.Next AppState)` is such an action. It depends if the configuration has a default command registered, otherwise focuses (read makes visible) the editor to type in a command.

Another example: `switchComposeEditor :: Action 'Threads 'ListOfThreads AppState` is an action which, depending if a composition is underway, goes back to the compose editor, otherwise does nothing. It could otherwise be asking for composing a new mail.

My point being is, that these actions are all or nothing actions which can't really be chained like most of others. I think it would be great if we could split these actions based on a condition.

**Describe the solution you'd like**
I haven't thought about this thoroughly and what implications it has. At this point, the actions don't have access to any form of context, so using Haskells `if then else` will probably not work when registering the `Action`.

**Describe alternatives you've considered**
 Maybe we need a `ConditionalAction` which can accept two individual actions? I don't know.

**Additional context**
See examples

frasertweedale:
So, turns out you can kinda sorta get a Monad instance, if you generalise the Action type just a bit.  See https://github.com/purebred-mua/purebred/commit/871de4a13a34b8bf1dea2e05f7bb394e745dde54.  That commit does not include a Monad instance, but the instance is straightforward (although it would violate the `(<*>) = ap` law).  This would give choice over "next action" based on result of current action.

It was nice to see the Profunctor and Category instances shake out of such a small change in the type.

All that said, I am not sure it is the right approach.  First, I am not comfortable with the lawless Monad instance.  Second, I am simultaneously investigating a different change to the Action type, to use `StateT AppState (EventM Name) a` instead of `AppState -> EventM Name a`.  This change will allow some nice cleanups and easier way to set/read state from within actions.  The type will still admit Functor and Applicative instances, I think, as well as Monad with the same caveat as above.  This change is a large refactor so I'll just chip away at it and then we can see if we like the result or not.

frasertweedale:
Can we close this now that actions use `MonadState AppState m`?

romanofski:
Yeah lets close it.

PR #361, opened 2020-01-30
Feature/314 forwarding

romanofski:


src/UI/Actions.hs
```diff
@@ -1486,16 +1516,14 @@ editAttachment s =
...
   -> L.List Name MIMEMessage
   -> L.List Name MIMEMessage
-upsertPart charsets newPart l =
+upsertPart newPart l =
   case L.listSelectedElement l of
     Nothing -> L.listInsert 0 newPart l
     Just (_, part) ->
-      if view (headers . contentDisposition . folded . filename charsets) part
-          == view (headers . contentDisposition . folded . filename charsets) newPart
+      if view headers part == view headers newPart
```
frasertweedale: So, now `upsertPart` will insert a new part of any headers changed.  I'm ok with that.  But it would help if the behaviour of the function were fully elaborated in the haddock.

frasertweedale: On further thought, although it's not something to tackle in this commit, I think we should take a different approach: we should store more state before we go to edit a (new or existing) part, so that we know what to do when editing returns, i.e.:

- if it is new, note that in the state
- if editing existing, store a "cursor" in the state, so that we know exactly what to update

That will allow e.g. editing of headers (which is how I like to write my mail - with all the headers there in the file being edited).

As I say, it is not for this commit.  I have always found `upsertPart` a rather confusing function.  And it turns out that the current approach is inflexible too.

frasertweedale: Maybe we can even pass a not-quite-saturated list update action along through the action, so we don't have to add anything in the application state.  Would be quite a refactor though, with updates to the `Action` data type and the `chain` behaviour.  Could be fun though :)

romanofski: I like the last approach, since you can implement what you mentioned in your second comment. I agree that the current approach is rather wonky :/ How much do you think does this idea relate to #294 ?

frasertweedale: I'll leave comments over in #33 294.

PR #363, opened 2020-02-03
instance Ixed (V a)

frasertweedale:


src/Purebred/LazyVector.hs
```diff
@@ -97,3 +100,16 @@ splitAt i (V chunks) = case chunks of
     | otherwise ->
       let (V h', V t') = splitAt (i - length h) (V t)
       in (V (h : h'), V t')
+
+type instance Index (V a) = Int
+type instance IxValue (V a) = a
+
+instance Ixed (V a) where
+  ix n = splitIso (n + 1) . _1 . _VIso . _last . _last
```
romanofski: Sorry for the dumb question, but I have to ask: you split a chunk? at the next item. That gives you two chunks. You pick the first chunk and grab the last item which is now the item you want by the given index. Does this iso also cause chunk fragmentation or is this only a computed "property"?

romanofski:
Thank you so much mate! I'll check it out tonight.

src/Purebred/LazyVector.hs
frasertweedale: Not a dumb question at all.  Yes, it can cause chunk fragmentation.  If it turns out to be a problem we can add defragmentation heuristics in a future patch.

frasertweedale: To clarify, it's only going to fragment the "source" value if you write back to it through the Iso (and therefore by the `ix` `Traversal`).  If only used to read the target, then of course the original `V` value remains unchanged.

romanofski: Aaah yep yep, the last comment explains what I was wondering about with "computed property". Cheers for clarifying :)

romanofski:
I'd like to merge this, the only nitpick I get from GHC is:

```
src/Purebred/LazyVector.hs:111:14: warning: [-Wname-shadowing]
    This binding for ‘n’ shadows the existing binding
      bound at src/Purebred/LazyVector.hs:108:6
    |
111 |     splitIso n = iso (splitAt n) (uncurry (<>))
```

Happy to fix it, but since you already authored I thought you may want to know, fix it up and we merge :)

frasertweedale:
@romanofski you can fix it, probably just replace 'n' with 'i'

PR #364, opened 2020-02-06
implement debug log

frasertweedale:
```
0a2e7b3 (Fraser Tweedale, 4 minutes ago)
   log mode and focus changes

   This is what the log system was born to do.  Reveal the deep mysteries of
   how purebred's UI works at runtime, because I can't understand the code :)

d4fc29a (Fraser Tweedale, 9 minutes ago)
   add --debug=FILE option

   Add an option to write debug output to a file.  No logging is actually done
   (yet).  No timestamps, levels or other formatting
   (yet).  This was just hacked up in a jiffy and if we want something more
   sophisticated, we should evaluate established solutions e.g. co-log.

   The log is stored as a sink action (Text -> IO ()) in the AppState. If no
   log file is specified via the --debug=FILE option, this is a no-op. 
   Otherwise we create a TQueue, the sink puts messages in the queue, and we
   spawn a thread to read from the queue and write the messages to a file.  We
   don't have any special handling of IO exceptions and in fact, I don't even
   know what happens if e.g. the file isn't writable, it's a normal file,
   disappears during use, etc. I guess I'll cross those bridges later.
```

### DIFF
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -140,15 +140,20 @@ module Purebred (
 import UI.App (theApp, initialState)
 
 import qualified Config.Dyre as Dyre
+import Control.Concurrent (forkIO)
+import Control.Concurrent.STM (atomically, newTQueueIO, readTQueue, writeTQueue)
 import qualified Control.DeepSeq
-import Control.Monad ((>=>), void)
+import Control.Monad ((>=>), forever, void)
 import Options.Applicative hiding (str)
 import qualified Options.Applicative.Builder as Builder
 import Data.List (elemIndex, isInfixOf, isPrefixOf)
 import System.Environment (lookupEnv)
 import System.FilePath (dropTrailingPathSeparator, joinPath, splitPath)
 import System.FilePath.Posix ((</>))
+import System.IO (BufferMode(LineBuffering), IOMode(AppendMode), hSetBuffering, openFile)
 import System.Random (RandomGen, getStdGen, randomRs)
+import qualified Data.Text as T
+import qualified Data.Text.IO as T
 import Data.Version (showVersion)
 import Paths_purebred (version, getLibDir)
 
@@ -165,7 +170,7 @@ import Error
 import qualified Graphics.Vty
 import Graphics.Vty.Attributes
 import Graphics.Vty.Input.Events (Event(..), Key(..), Modifier(..))
-import Brick.BChan (BChan, newBChan)
+import Brick.BChan (newBChan)
 import Brick.Main (customMain)
 import Brick.Types (Next)
 import Brick.Util (on, fg, bg)
@@ -177,6 +182,7 @@ import Data.MIME (Mailbox(..), AddrSpec(..), Domain(..))
 data AppConfig = AppConfig
     { databaseFilepath :: Maybe String
     , searchOverride :: Maybe String
+    , debugFile :: Maybe FilePath
     }
 
 appconfig :: Parser AppConfig
@@ -193,6 +199,13 @@ appconfig = AppConfig
       ( long "search"
       <> metavar "SEARCH-TERM"
       <> help "Override the initial notmuch search"
+      )
+    )
+  <*> optional
+    ( Builder.option Builder.str
+      ( long "debug"
+      <> metavar "FILE"
+      <> help "Write debug information to FILE"
       )
     )
   <* Builder.infoOption versionString
@@ -234,7 +247,18 @@ launch cfg = do
   --
   bchan <- newBChan 32
 
-  cfg' <- processConfig (bchan, b) (pre cfg)
+  -- Create log sink.
+  logSink <- case debugFile opts of
+    Nothing -> pure $ \_ -> pure ()
+    Just fp -> do
+      h <- openFile fp AppendMode
+      hSetBuffering h LineBuffering
+      T.hPutStrLn h $ T.pack "Opened log file"
+      q <- newTQueueIO
+      _ <- forkIO $ forever $ atomically (readTQueue q) >>= T.hPutStrLn h
+      pure $ atomically . writeTQueue q
+
+  cfg' <- processConfig (bchan, b, logSink) (pre cfg)
 
   s <- initialState cfg'
   let buildVty = Graphics.Vty.mkVty Graphics.Vty.defaultConfig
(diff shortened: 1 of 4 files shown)


src/UI/Actions.hs
```diff
@@ -860,7 +865,14 @@ abort = Action ["cancel"] (reset (Proxy :: Proxy v) (Proxy :: Proxy a))
 focus :: forall a v. (HasViewName v, HasName a, Focusable v a) => Action v a AppState
 focus = Action
   ["switch mode to " <> T.pack (show (name (Proxy :: Proxy a)))]
-  (switchFocus (Proxy :: Proxy v) (Proxy :: Proxy a))
+  (\s -> do
+    liftIO . view (asConfig . confLogSink) s . T.pack $
+      "switchFocus "
+        <> show (viewname (Proxy :: Proxy v)) <> " "
+        <> show (name (Proxy :: Proxy a))
```
endgame: Type applications? `Proxy @a`?

src/UI/Actions.hs
```diff
@@ -860,7 +865,14 @@ abort = Action ["cancel"] (reset (Proxy :: Proxy v) (Proxy :: Proxy a))
 focus :: forall a v. (HasViewName v, HasName a, Focusable v a) => Action v a AppState
 focus = Action
   ["switch mode to " <> T.pack (show (name (Proxy :: Proxy a)))]
-  (switchFocus (Proxy :: Proxy v) (Proxy :: Proxy a))
+  (\s -> do
+    liftIO . view (asConfig . confLogSink) s . T.pack $
+      "switchFocus "
```
endgame: When I'm going a big concat I tend to `mconcat` a list. Maybe you might like how that looks?

src/UI/Actions.hs
```diff
@@ -832,17 +832,22 @@ chain (Action d1 f1) (Action d2 f2) = Action (d1 <> d2) (f1 >=> f2)
...
-      . over (asViews . vsFocusedView) (Brick.focusSetCurrent (viewname (Proxy :: Proxy v')))
-      . set (asViews . vsViews . at (viewname (Proxy :: Proxy v')) . _Just . vFocus) (name (Proxy :: Proxy ctx'))
+    switchMode s = do
+      liftIO . view (asConfig . confLogSink) s . T.pack $
+        "chain' "
+          <> show (viewname (Proxy :: Proxy v)) <> "/" <> show (name (Proxy :: Proxy ctx)) <> " -> "
+          <> show (viewname (Proxy :: Proxy v')) <> "/" <> show (name (Proxy :: Proxy ctx'))
+      pure $ s &
+        transitionHook (Proxy :: Proxy v) (Proxy :: Proxy v')
+        . over (asViews . vsFocusedView) (Brick.focusSetCurrent (viewname (Proxy :: Proxy v')))
```
romanofski: I welcome this, because this could make quite a bit of code redundant I've been dragging around. However there is an additional bit missing, which set's the widgets state in the view to visible. Since the widgets sit in a list of lists, the code would need an O(n) operation to find the widget's "Layer" or ... well use a better data structure. Have a look at some of the `Focusable` instances. Not sure whether you're keen ...

src/Types.hs
```diff
@@ -354,6 +354,11 @@ fbSearchPathKeybindings = lens _fbSearchPathKeybindings (\cv x -> cv { _fbSearch
 fbHomePath :: Lens (FileBrowserSettings a) (FileBrowserSettings a') a a'
 fbHomePath = lens _fbHomePath (\s a -> s { _fbHomePath = a })
 
+data Delay
```
romanofski: This looks like a stray addition from https://github.com/purebred-mua/purebred/commit/44946bfe62eb8288fd2cf45ef43200de59ca0a81 ?

src/Purebred.hs
```diff
@@ -234,7 +247,18 @@ launch cfg = do
...
   bchan <- newBChan 32
 
-  cfg' <- processConfig (bchan, b) (pre cfg)
+  -- Create log sink.
+  logSink <- case debugFile opts of
+    Nothing -> pure $ \_ -> pure ()
+    Just fp -> do
+      h <- openFile fp AppendMode
+      hSetBuffering h LineBuffering
+      T.hPutStrLn h $ T.pack "Opened log file"
```
romanofski: Could we put this into a general logging function? The reason why I'm asking is that it would be really good to also have timestamps next to the entries. I would love to add that to our tests and have it shown at the end of the test session. Without timestamps it'll be really hard to figure out when what happened.

frasertweedale: save for another PR.  feel free to tackle it!

src/Types.hs
frasertweedale: nah I just moved it.  it was in the way :)  see removed lines below.

src/UI/Actions.hs
frasertweedale: What code will it make redundant?

Any enhancements/additions can wait for another PR.  I don't did the bare minimum I needed, but you are the UI author and have a better idea what would be worth logging.

romanofski: Fair enough!

src/Types.hs
romanofski: Ah dang!

src/Purebred.hs
romanofski: Alright

src/UI/Actions.hs
frasertweedale: all reduces the same program :)  it's just an aesthetic thing.

If I want to go wild I'd sooner switch to something using `Data.Text.Builder` so that the logging framework can stream it straight into the file (well, maybe it already fuses!  I haven't read the core yet...)

src/UI/Actions.hs
frasertweedale: I thought about it.  I'm not against it but I didn't want to either (a) refactor the classes in this commit or (b) apply the functions to `undefined` as the (still currently required) argument.  Feel free to open a PR to do the refactor (after this one gets merged).

romanofski: TIL about type applications. Why would type applications make sense here? I've briefly browsed through the proposal and the way I understood it is to be more explicit where ambiguity exists. Does the ambiguity here come with the use of `show`?

frasertweedale: @endgame ah, nevermind, I understand what you were suggesting now.  Besides, there's no way to get rid of the proxy argument, otherwise GHC will not accept the class definition :)  But... it can wait for another PR.

frasertweedale: @romanofski it's nothing to do with `show`.  Just makes constructing `Proxy` values a bit cleaner, among other uses.
```
λ> :set -XTypeApplications
λ> import Data.Proxy
λ> :t Proxy @Int
Proxy @Int :: Proxy Int
```

PR #365, opened 2020-02-07
refactor Actions to use MonadState

frasertweedale:
"Make the whole country seethe with a high-pitched campaign for producing greenhouse vegetables!"

```
f87f82e (Fraser Tweedale, 33 minutes ago)
   use -XTypeApplications for 'focus' in keybindings

   Wanting to relax the return type of the 'focus' action from
   'AppState' to '()' revealed an opportunity to improve usability.  In 
   keybindings, when 'focus' is used we currently use an explicit, fully
   saturated type signature to determine the Focusable instance. This includes
   the 'Action' constructor and the 'AppState' return type, which are both
   redundant (`chain` and `chain'` both ignore the Action return type as a
   result of the StateT refactor).

   Therefore we can simplify the definition of keybindings that use
   'focus' by using -XTypeApplications, available in GHC since 8.0.

654f2c8 (Fraser Tweedale, 54 minutes ago)
   refactor invokeEditor' and editAttachment

   Refactor these functions to use 'MonadState'.  In the case of
   'editAttachment', a lot of the behaviour was subsumed by an application of
   'selectedItemHelper'.

e58e41d (Fraser Tweedale, 75 minutes ago)
   fix instance Focusable 'Threads 'ComposeFrom

   The old implementation (which was factored out to a top-level function
   'focusComposeFrom' contained the expression:

     if nullOf (asCompose . cAttachments) s
        then ...

   This was erroneously refacted as:

     l <- use (asCompose . cAttachments)
    if null l
      then ...

   The latter is testing whether the attachment list is empty, whereas the
   former is testing whether the optic (asCompose . cAttachments) has zero
   targets.  In the case of an Iso, Lens or Getter, there is always exactly
   one target, hence 'nullOf' always returns 'False'! The true branch of the
   old implementation is unreachable, but I introduced a bug (that made it
   reachable) during the refactor.

   To fix, remove the conditional and replace the whole conditional with the
   false branch.

c278f3d (Fraser Tweedale, 75 minutes ago)
   refactor Action to use StateT

   Using 'StateT AppState (EventM n)' for 'Action' definitions improves the
   readability (and writeability) of actions.  It also increases the
   flexibility (or abstractness) of what data types actions return.

   To get all this to work we had to add some orphan 'MonadThrow',
   'MonadCatch' and 'MonadMask' instances for '(EventM n)'.  I hope to land
   these upstream in brick.

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -330,7 +330,7 @@ unhide = setViewState Visible
 
 setViewState :: (MonadState AppState m) => ViewState -> ViewName -> Int -> Name -> m ()
 setViewState v n i m =
-  assign (asViews . vsViews . at n . _Just . vLayers . ix i . ix m . veState) v
+  assign (asViews . vsViews . ix n . vLayers . ix i . ix m . veState) v
 
 instance Completable 'ComposeTo where
   complete _ = do
@@ -523,33 +523,33 @@ instance Focusable 'Threads 'ListOfThreads where
 instance Focusable 'ViewMail 'ManageMailTagsEditor where
   switchFocus _ _ = do
     unhide ViewMail 0 ManageMailTagsEditor
-    assign (asViews . vsViews . at ViewMail . _Just . vFocus) ManageMailTagsEditor
+    assign (asViews . vsViews . ix ViewMail . vFocus) ManageMailTagsEditor
 
 instance Focusable 'ViewMail 'ScrollingMailView where
-  switchFocus _ _ = assign (asViews. vsViews . at ViewMail . _Just . vFocus) ScrollingMailView
+  switchFocus _ _ = assign (asViews . vsViews . ix ViewMail . vFocus) ScrollingMailView
 
 instance Focusable 'ViewMail 'ScrollingMailViewFindWordEditor where
   switchFocus _ _ = do
     modifying (asMailView . mvFindWordEditor . E.editContentsL) clearZipper
-    assign (asViews. vsViews . at ViewMail . _Just . vFocus) ScrollingMailViewFindWordEditor
+    assign (asViews. vsViews . ix ViewMail . vFocus) ScrollingMailViewFindWordEditor
     unhide ViewMail 0 ScrollingMailViewFindWordEditor
 
 instance Focusable 'ViewMail 'ListOfMails where
-  switchFocus _ _ = assign (asViews . vsViews . at ViewMail . _Just . vFocus) ListOfMails
+  switchFocus _ _ = assign (asViews . vsViews . ix ViewMail . vFocus) ListOfMails
 
 instance Focusable 'ViewMail 'MailListOfAttachments where
   switchFocus _ _ = do
-    assign (asViews . vsViews . at ViewMail . _Just . vFocus) MailListOfAttachments
+    assign (asViews . vsViews . ix ViewMail . vFocus) MailListOfAttachments
     unhide ViewMail 0 MailListOfAttachments
 
 instance Focusable 'ViewMail 'MailAttachmentOpenWithEditor where
   switchFocus _ _ = do
-    assign (asViews . vsViews . at ViewMail . _Just . vFocus) MailAttachmentOpenWithEditor
+    assign (asViews . vsViews . ix ViewMail . vFocus) MailAttachmentOpenWithEditor
     unhide ViewMail 0 MailAttachmentOpenWithEditor
 
 instance Focusable 'ViewMail 'MailAttachmentPipeToEditor where
   switchFocus _ _ = do
-    assign (asViews . vsViews . at ViewMail . _Just . vFocus) MailAttachmentPipeToEditor
+    assign (asViews . vsViews . ix ViewMail . vFocus) MailAttachmentPipeToEditor
     unhide ViewMail 0 MailAttachmentPipeToEditor
 
 instance Focusable 'ViewMail 'SaveToDiskPathEditor where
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: src/UI/Actions.hs [hunk 1]
```diff
@@ -1515,13 +1513,11 @@ pipeCommand' cmd
...
 
-editAttachment :: AppState -> IO AppState
-editAttachment s =
-    case L.listSelectedElement $ view (asCompose . cAttachments) s of
-        Nothing -> pure $ setError (GenericError "No file selected to edit") s
-        Just (_, m) -> case preview (headers . contentDisposition . folded . dispositionType) m of
-          (Just Inline) -> invokeEditor' s
-          _ -> pure $ setError (GenericError "Not implemented. See #182") s
+editAttachment :: (MonadState AppState m, MonadIO m, MonadMask m) => m ()
+editAttachment = selectedItemHelper (asCompose . cAttachments) $ \m -> do
```
romanofski: Full disclosure: I really hate the `selectedItemHelper` function ;) I've come to hate it while hacking on #5 since the refactoring is fiddly ...

src/UI/Actions.hs
```diff
@@ -287,44 +299,52 @@ instance HasList 'ListOfFiles where
...
     case getEditorTagOps (Proxy :: Proxy 'ManageMailTagsEditor) s of
         Left err -> pure $ setError err s
-        Right ops' -> over (asMailIndex . miListOfThreads) (L.listModify (Notmuch.tagItem ops'))
-                      <$> selectedItemHelper (asMailIndex . miListOfMails) s (manageMailTags s ops')
+        Right ops' -> flip execStateT s $ do
+          modifying (asMailIndex . miListOfThreads) (L.listModify (Notmuch.tagItem ops'))
+          selectedItemHelper (asMailIndex . miListOfMails) (manageMailTags ops')
+
+
+hide, unhide :: (MonadState AppState m) => ViewName -> Int -> Name -> m ()
```
romanofski: Ah nice. Well thank you! Although if it isn't too much effort I'd put them in `UI.Views`?

src/UI/Actions.hs
```diff
@@ -737,44 +748,47 @@ instance HasViewName 'FileBrowser where
...
-continue = Action mempty Brick.continue
+continue = Action mempty (get >>= lift . Brick.continue)
 
 -- | Suspends Purebred and invokes the configured editor.
 --
 invokeEditor :: Action v ctx (T.Next AppState)
-invokeEditor = Action ["invoke external editor"] (Brick.suspendAndResume . liftIO . invokeEditor')
+invokeEditor = Action ["invoke external editor"]
+  (lift . Brick.suspendAndResume . liftIO . invokeEditor' =<< get)
+  -- TODO abstract invokeEditor'
```
romanofski: Can we put this TODO into an issue pls?

src/UI/Actions.hs
```diff
@@ -862,316 +884,333 @@ abort = Action ["cancel"] (reset (Proxy :: Proxy v) (Proxy :: Proxy a))
...
-                          (view (asFileBrowser . fbEntries . L.listSelectedL) s)
+    , _aAction = use (asFileBrowser . fbEntries . L.listSelectedL) >>= traverse_ f
     }
+    where
+      f i = modifying (asFileBrowser . fbEntries . L.listElementsL . ix i . _1) not
 
 -- | Delete an attachment from a mail currently being composed.
 --
-delete :: Action 'ComposeView 'ComposeListOfAttachments AppState
+-- TODO: could this be generalised over a type class, e.g.:
```
romanofski: This one as well into an issue with an easy fix?

src/UI/Actions.hs
frasertweedale: It's done in a later commit in this PR.

src/UI/Actions.hs
frasertweedale: Yes I'll file an issue for it.

src/UI/Actions.hs
frasertweedale: not right now.  there's already enough in this refactor :D

src/UI/Actions.hs
frasertweedale: I refactored `selectedItemHelper`!  What do you think of the new and improved version?  I find it reasonably pleasant and intuitive now that it is `MonadState` based.

romanofski:
Merge once CI is green!

frasertweedale:
@romanofski do you know what is going on with the nix job?

romanofski:
One of the haddock strings is cactus:

```
src/UI/Actions.hs:1156:9: error:
    parse error on input ‘-- | Construct the full path to the attachment. The file browser only
        -- lists the file names (otherwise we wouldn't have the ability to
        -- display the full paths for deeper file hirarchies). However when
        -- attaching the file, we need the full paths so we can find, edit and
        -- update the attachments later.’
     |
1156 |         -- | Construct the full path to the attachment. The file browser only
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

romanofski:
I'll fix it ...

src/UI/Actions.hs
romanofski: I like it! Specifically the part that it doesn't use `AppState -> AppState` as a parameter any more :)

frasertweedale:
@romanofski don't we run haddock in the other builds too?  Is there a different version of haddock on nix?

romanofski:
Heh actually we do and they do fail: https://travis-ci.org/purebred-mua/purebred/jobs/647339453?utm_medium=notification&utm_source=github_status (scroll all the way down). The exit status must be swallowed tho ...

frasertweedale:
@romanofski anyhow thanks for the fixup!

My brick PR to add `MonadMask` instance for `EventM` got merged already, so as soon as that's released we can bump with lower bound and drop the orphan instances.  But that can happen later.

frasertweedale:
@romanofski ok, that's a CI bug then.  I'll file it.

romanofski:
There also seems to be a version difference because this one: https://travis-ci.org/purebred-mua/purebred/jobs/647519007?utm_medium=notification&utm_source=github_status has a parse error in `Type.hs`, while the other builds seem fine.

frasertweedale:
@romanofski if you're happy, merge it.  I've got to head out shortly and won't be online all day :)

romanofski:
Yeah lets merge. CI says green. For the non-green issues we can look separately.

PR #369, opened 2020-02-10
Make cc bcc editable

### LINKED ISSUE #290: CC and BCC fields need to be editable
**Is your feature request related to a problem? Please describe.**
When sending mail, I can not edit CC or BCC fields.

**Describe the solution you'd like**
In the compose editor, we need to show two new fields: CC and BCC. The state will need to be updated too.

**Describe alternatives you've considered**
None

**Additional context**
None

voanhduy1512:
Fix #290


voanhduy1512:
Not totally related to this PR but i am not sure how to run UAT test in local development. I try to read ci config to figure it out but it seems a bit difficult to understand. Can you help me with the instruction on how to run it

romanofski:
@voanhduy1512 This is looking like it's going in the right direction!! Very nice! I won't have time today to review unfortunately. I hope you can wait a day or two.
Just to give you a pointer in regards to the user acceptance tests: I have documented a few hints on how to get started in our [HACKING](https://github.com/purebred-mua/purebred/blob/master/HACKING) document which you can find in the root of the repository. Perhaps that may get you going in the mean time.

I have the suspicion that not many people find this document. @frasertweedale Perhaps we should merge HACKING with the README.md so it's all basically on the front page?

voanhduy1512:
No worries, please review whenever you have free time.

src/UI/ComposeEditor/Main.hs
```diff
@@ -51,7 +51,9 @@ attachmentsEditor s =
     in attachmentsHeader <=> attachmentsList
 
 drawHeaders :: AppState -> Widget Name
-drawHeaders s = padBottom (Pad 1) $ foldr (drawTableRows s) (txt T.empty) [ComposeSubject, ComposeTo, ComposeFrom]
+drawHeaders s = 
+  let headers' = [ComposeSubject, ComposeTo, ComposeFrom, ComposeCc, ComposeBcc]
```
romanofski: At the moment the fields are rendered like so:
screenshot

I suspect if you stick the `CC` and `BCC` `Name`s between `ComposeSubject` and `ComposeTo` they will be rendered below the `To:` line. I think that would be the better place since they're not as often used as `To` and `From`

test/TestUserAcceptance.hs
```diff
@@ -570,6 +570,34 @@ testEditingMailHeaders = purebredTmuxSession "user can edit mail headers" $
     sendKeys ", testuser@foo.test\r" (Substring "To: user@to.test, testuser@foo.test")
       >>= assertRegex lastLineIsStatusLine
 
+    step "user can add cc header"
```
romanofski: Cool! Did you find it on your own? I think that's great and would leave it.

romanofski: The config we're using for the user acceptance tests writes the mail out to a file instead of actually sending it. Some tests (e.g. "forwards mail successfully") analyses the "sent" mail by opening it up and asserting that certain sub strings are found. It's very crude but we haven't had a need for something more elaborate.

I wish we would have put the same procedure into the 'sending mail successfully'. It currently only asserts that the mail can be parsed.

Anyway, what I'm getting at is, that it may be good to also check whether the mail contains what we edit in the `CC` and `BCC` fields. If you feel keen to add it, perhaps add entering `BCC` and `CC` addresses to the "forwards mail successfully" test case. 

If you don't feel like messing around more with it it's fine too and we add it in a separate issue.

Let me know if you need any help with it.

romanofski:
@voanhduy1512 Let me know if you have any troubles with the patch :) It's 99% there.

test/TestUserAcceptance.hs
voanhduy1512: Sorry for the delay update, it should be fixed now

romanofski:
\o/ Nice!!

PR #374, opened 2020-02-18
send mail / save draft: cleanup and perf improvement

frasertweedale:
```
aa18c96 (Fraser Tweedale, 3 hours ago)
   avoid construction of strict ByteStrings

   When sending or saving mail, we currently render the mail into a strict
   ByteString.  This could be quite an expensive operation when all we really
   need to do is send it to sendmail(1) stdin or to a file.  Change the
   cvSendMailCommand to accept a Builder instead of a strict ByteString,
   allowing a more efficient implementation, and chase the changes.

   Also remove the indexMail function.  It should not be the responsibility of
   the Notmuch subroutines to actually write the mail to disk (and pushing
   that responsibility to the caller allows the caller to use the most
   efficient means of writing the file).

c874068 (Fraser Tweedale, 3 hours ago)
   refactor send mail and safe draft actions

   Refactor the send mail and save draft actions to use MonadState.
```

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -124,8 +124,7 @@ import Data.List.NonEmpty (NonEmpty(..))
 import System.FilePath (takeDirectory, (</>))
 import qualified Data.Vector as Vector
 import Prelude hiding (readFile, unlines)
-import Data.Foldable (traverse_)
-import Data.Functor (($>))
+import Data.Foldable (toList, traverse_)
 import Control.Lens
        (_Just, to, at, ix, _1, _2, toListOf, traverse, traversed, has, snoc,
         filtered, set, over, preview, view, (&), firstOf, non,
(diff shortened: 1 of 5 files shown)


src/UI/Actions.hs
```diff
@@ -1381,17 +1387,12 @@ buildMail s = do
...
-    let cmd = view (asConfig . confComposeView . cvSendMailCmd) s
-        defMailboxes = view (asConfig . confComposeView . cvIdentities) s
-    catch
-        (cmd m $> (s
-         & set asCompose (initialCompose defMailboxes)
-         . set (asConfig . confBoundary) l'))
-        (\e ->
-              let err = show (e :: IOException)
-              in pure $ s & setError (SendMailError err))
+trySendAndCatch :: (MonadState AppState m, MonadIO m, MonadCatch m) => B.Builder -> m ()
```
romanofski: I have the feeling that the above comment can be deleted with the refactoring? Seems like this change is solving the problem I had back then?

src/UI/Actions.hs
```diff
@@ -1534,27 +1535,23 @@ manageThreadTags ops t = do
...
+  modify clearMailComposition
+
+keepDraft :: (MonadMask m, MonadState AppState m, MonadIO m) => m ()
+keepDraft = buildMail $ \bs -> do
+  maildir <- use (asConfig . confNotmuch . nmDatabase)
+  draftTag <- use (asConfig . confNotmuch . nmDraftTag)
+  runExceptT ( do
+      fp <- createDraftFilePath maildir
+      tryIO $ LB.writeFile fp (B.toLazyByteString bs)
+      Notmuch.indexFilePath maildir fp [draftTag] )
```
romanofski: Ah nice. Good stuff!

romanofski:
I guess the only thing is the comment which seems to be redundant now, unless I missed something. If still valid merge ahead!

frasertweedale:
Deleted the comment.  Merging.

PR #377, opened 2020-02-29
Log with a timestamp

romanofski:
This adds a new logging module providing a simple logging wrapper adding
a timestamp to the log entry.

### DIFF
--- a/src/Purebred/System/Logging.hs
+++ b/src/Purebred/System/Logging.hs
@@ -1,0 +1,50 @@
+-- This file is part of purebred
+-- Copyright (C) 2020 Fraser Tweedale and Róman Joost
+--
+-- purebred is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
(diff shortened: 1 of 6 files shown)


romanofski:
Almost wondering whether we should add an acceptance test. Not really an end user feature so I thought meh...

src/Purebred/System/Logging.hs
```diff
@@ -0,0 +1,50 @@
...
+      hSetBuffering h LineBuffering
+      q <- newTQueueIO
+      _ <- forkIO $ forever $ atomically (readTQueue q) >>= log_ h
+      pure $ atomically . writeTQueue q
+
+log_ :: Handle -> T.Text -> IO ()
+log_ h logtext = do
+  dateTimeNow <- getZonedTime
+  let timestamp = T.pack $ formatTime defaultTimeLocale "%c" dateTimeNow
+  T.hPutStrLn h $ timestamp <> " " <> logtext
```
frasertweedale: Another suggestion: switch to `Data.Text.Lazy.Text` for efficiency.  Can be done in a separate commit.

romanofski: Ah yep yep.. Good idea. Will update it.

PR #380, opened 2020-03-29
Feature/5 marking for bulk actions

romanofski:
Oof this has now become quite large, but worth it I guess. I'm anticipating that you might not have the time instantly to go through all of that. The way how currently toggled list items are rendered sucks a bit and I may investigate further if that can be improved.
However I feel fine to merge it as it is.
```
commit a15e987b45c7c74e5dacde5bb6067047e9789e19
Date:   Sun Mar 29 15:42:40 2020 +1000

    Re-use HasToggleablelist functions instead of utils
    
    This is a small refactoring to re-use the HasToggleablelist functions
    instead of the utility functions only used for the file browser. This
    doesn't make the attachment creation generic, since all of this code is
    verys specific to the file browser anyways.

commit 6182328255718612f45b8ac90bf3d181d6018722
Date:   Sun Mar 29 12:24:09 2020 +1000

    Test bulk actions with additional acceptance tests

commit 30fa4a15095eebc8be756a8ca3285a2d2ed5feb8
Date:   Sun Mar 29 07:20:21 2020 +1000

    Use a better word than "Tagged"
    
    I had used "Tag", because mutt is using "Tag". The word tagging here has
    a different context however. In mutt it used to be to perform bulk
    actions on the tagged list items, but with the introduction of notmuch,
    tagging can also mean the labeling of mails with "tags".
    
    Instead of using "Tagged", use the word "Marked" since we're marking
    these list items for further action.

commit 43de3888192cedb6a4913e87594bc902d95e3998
Date:   Sat Mar 28 16:47:43 2020 +1000

    Operate on toggled *or* selected list items
    
    This adds an additional helper for lists supporting toggling of it's
    items. It is similar in nature to our existing `selectedItemHelper`.
    
    It really is a workaround for Brick's List, since it can not encapsulate
    a marked/toggled state for it's list items. This function allows to
    operate on toggled or selected items.

commit 5e14cb4d43aad4fe705b934da33d873e6d26d097
Date:   Sat Mar 28 16:29:04 2020 +1000

    Use `toggle` to indicate a selected state
    
    We currently use selected and toggled ambiguosly in the code. To make it
    clear: refer to a list item which is toggled for batch editing as
    `toggled` and a list item which is selected by the cursor as: selected.

commit 3702b535a5fed8dc79baa3986ce7d1050d0a54b9
Date:   Sat Mar 28 10:34:19 2020 +1000

    Raise an error if setTags is called on unsupported widget
    
    If we call `setTags` on a widget we have not registered a handler for,
    Purebred would simply assume that the only widget would be the list of
    threads.
    
    However when adding new widgets, this can lead to confusing
    behaviour for new contributors since the handler is silently operating
    on the list of threads.

commit 68b2cf2f01356e0cc0c231adabb2a4dd63418675
Date:   Sat Mar 28 10:33:39 2020 +1000

    Remove selectAll for now
    
    At this point we do not want to support toggling all items in a list due
    to performance concerns.

commit b40697fee779705b3835e1a9357fdcc990fa782d
Date:   Sun Mar 22 17:04:59 2020 +1000

    derive HasSelectableItemList for any context with (Bool, a) element type

commit e11d7bc1fc2414aaddf4c8312074dae7afd0906b
Date:   Sun Mar 22 13:09:27 2020 +1000

    Provide a lens to operate on toggled items

commit 15cfcf41a4d24541ecf2464ce6832a98182880aa
Date:   Sun Feb 16 14:41:00 2020 +1000

    Improve help for setTags action
    
    The 'apply given tags' isn't very helpful, particularly when we have a
    look at our help page in the UI:
    
         a - apply given tags
         S - apply given tags
    
    Both are distinct actions and the 'what' is missing here. This adds at
    least the show instances of the tag operations to make this part of the
    UI more helpful.

commit 583054f1b13c2604588aacfa2da0b5c977f87750
Date:   Thu Feb 6 19:41:07 2020 +1000

    Bulk actions
    
    This implements the support for bulk actions, currently limited to
    tagging only. The major part of this patch is the support for tagging
    multiple items (either mail or thread) from our Brick list widgets now
    able to hold a toggle state in the list item itself.
    
    Fixes https://github.com/purebred-mua/purebred/issues/5

commit 8c08052ac151bb59cacc6c551c270c9679ed3ef7
Date:   Thu Jan 30 20:12:26 2020 +1000

    Type alias to reflect marked list items
    
    This is a simple type alias which reflects list items which can be
    marked for bulk actions.
```

### DIFF
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -230,7 +230,7 @@ getThreads
   :: (MonadError Error m, MonadIO m)
   => T.Text
   -> NotmuchSettings FilePath
-  -> m (V (SelectableItem NotmuchThread))
+  -> m (V (Toggleable NotmuchThread))
 getThreads s settings =
   withDatabaseReadOnly (view nmDatabase settings) $
     flip Notmuch.query (Notmuch.Bare $ T.unpack s)
@@ -247,7 +247,7 @@ getThreadMessages
   :: (MonadError Error m, MonadIO m, Traversable t)
   => FilePath
   -> t NotmuchThread
-  -> m (Vec.Vector (SelectableItem NotmuchMail))
+  -> m (Vec.Vector (Toggleable NotmuchMail))
 getThreadMessages fp ts = withDatabaseReadOnly fp go
   where
     go db = do
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -48,10 +48,11 @@ module Storage.Notmuch (
   , withDatabase
   ) where
 
-import Control.Monad ((>=>), when)
+import Control.Monad ((<=<), (>=>), when)
 import Data.Function (on)
 import System.IO.Unsafe (unsafeInterleaveIO)
-import Data.Foldable (foldlM)
+import Data.Foldable (toList)
+import Data.Functor.Compose (Compose(..))
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Except (MonadError, throwError, ExceptT, runExceptT)
@@ -250,11 +251,11 @@ getThreadMessages
 getThreadMessages fp ts = withDatabaseReadOnly fp go
   where
     go db = do
-      msgs <- foldlM (allMessages db) [] ts
-      mails <- liftIO $ traverse messageToMail msgs
-      pure $ Vec.fromList $ fmap (False,) mails
-    allMessages db acc t =
-      mappend acc <$> (getThread db (view thId t) >>= Notmuch.messages)
+      msgs <-
+        Data.Functor.Compose.Compose  -- 'collapse' nested 't ([] a)'
+        <$> traverse (Notmuch.messages <=< getThread db . view thId) ts
+      mails <- traverse (liftIO . messageToMail) msgs
+      pure . Vec.fromList . toList $ fmap (False,) mails
 
 
 -- | retrieve a given thread from the notmuch database by it's id
--- a/src/Storage/Notmuch.hs
+++ b/src/Storage/Notmuch.hs
@@ -1,5 +1,5 @@
 -- This file is part of purebred
--- Copyright (C) 2017-2019 Róman Joost
+-- Copyright (C) 2017-2020 Róman Joost
 --
 -- purebred is free software: you can redistribute it and/or modify
 -- it under the terms of the GNU Affero General Public License as published by
@@ -17,6 +17,8 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TupleSections #-}
 
 module Storage.Notmuch (
     -- * Synopsis
@@ -49,6 +51,7 @@ module Storage.Notmuch (
 import Control.Monad ((>=>), when)
 import Data.Function (on)
 import System.IO.Unsafe (unsafeInterleaveIO)
+import Data.Foldable (foldlM)
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Except (MonadError, throwError, ExceptT, runExceptT)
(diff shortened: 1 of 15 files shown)


frasertweedale:
@romanofski something busted in `unit` test suite:
```
test/TestActions.hs:32:18: error:
    Module ‘UI.Utils’ does not export ‘selectedFiles’
   |
32 | import UI.Utils (selectedFiles)
   |                  ^^^^^^^^^^^^^
cabal: Failed to build test:unit from purebred-0.1.0.0.
```

src/Storage/Notmuch.hs
```diff
@@ -226,30 +229,34 @@ getThreads
...
-  where go db = do
-          msgs <- getThread db (view thId t) >>= Notmuch.messages
-          mails <- liftIO $ traverse messageToMail msgs
-          pure $ Vec.fromList mails
+  -> t NotmuchThread
+  -> m (Vec.Vector (Toggleable NotmuchMail))
+getThreadMessages fp ts = withDatabaseReadOnly fp go
+  where
+    go db = do
+      msgs <- foldlM (allMessages db) [] ts
```
frasertweedale: Why `foldlM` specifically?

Would something like
```haskell
msgs <- mconcat <$> traverse (Notmuch.messages <=< getThread db . view thId) ts
```

work?  I just find traversals and monoids much easier to read than folds.

*edit* you could even avoid the `mconcat` and just do another layer of `traverse` in the next line, e.g.:
```haskell
msgs <- traverse (Notmuch.messages <=< getThread db . view thId) ts
mails <- liftIO $ (traverse . traverse) messageToMail msgs
```

src/UI/Actions.hs
```diff
@@ -289,9 +290,51 @@ instance HasList 'MailListOfAttachments where
...
+  toggleE :: Proxy n -> E n -> E n
+  isToggledE :: Proxy n -> E n -> Bool
+  untoggleAll :: (MonadState AppState m) => Proxy n -> m ()
+  untoggleAll proxy = modifying (list proxy . traversed) (untoggleE proxy)
+  toggle :: Proxy n -> Int -> StateT AppState (T.EventM Name) ()
+  -- | Traversal of selected items.  NOT a valid Traversal unless
+  -- selected state is preserved
+  toggledItemsL :: Proxy n -> Traversal' AppState (E n)
+  toggledItemsL proxy = list proxy . traversed . filtered (isToggledE proxy)
+  toggledOrSelectedItemHelper ::
```
frasertweedale: could we move this out of the class definition?

src/UI/Actions.hs
```diff
@@ -1006,8 +1055,8 @@ displayThreadMails =
         -- Update the Application state with all mails found for the
         -- currently selected thread.
         dbpath <- use (asConfig . confNotmuch . nmDatabase)
-        selectedItemHelper (asMailIndex . miListOfThreads) $ \t ->
-          runExceptT (Notmuch.getThreadMessages dbpath t)
+        selectedItemHelper (asMailIndex . miListOfThreads) $ \(_, t) ->
+          runExceptT (Notmuch.getThreadMessages dbpath [t])
```
frasertweedale: nit: I'd prefer `Identity` here rather than `[]`.  Not because it's lower runtime overhead (though it is)
but because it more accurately expresses the intent, i.e. that there is always exactly one value in
this case.

src/UI/Actions.hs
```diff
@@ -1112,15 +1161,25 @@ toggleHeaders = Action
 -- | Apply given tag operations on the currently selected thread or
 -- mail.
 --
-setTags :: [TagOp] -> Action v ctx ()
+setTags :: forall v ctx. HasToggleableList ctx => [TagOp] -> Action v ctx ()
 setTags ops =
     Action
-    { _aDescription = ["apply given tags"]
+    { _aDescription = ["apply tag operations: " <> T.intercalate ", " (T.pack . show <$> ops) ]
```
frasertweedale: nice.

src/UI/Index/Keybindings.hs
```diff
@@ -39,6 +39,7 @@ browseThreadsKeybindings =
     , Keybinding (V.EvKey V.KUp []) (listUp `chain` continue)
     , Keybinding (V.EvKey (V.KChar 'G') []) (listJumpToEnd `chain` continue)
     , Keybinding (V.EvKey (V.KChar '1') []) (listJumpToStart `chain` continue)
+    , Keybinding (V.EvKey (V.KChar '*') []) (toggleListItem `chain` listDown `chain` continue)
```
frasertweedale: So happy our UI is configurable enough that I can easily override this behaviour to *not* `listDown` :)

src/UI/Status/Main.hs
```diff
@@ -108,15 +108,26 @@ renderStatusbar :: WithContext w => w -> AppState -> Widget Name
...
+  , renderToggled s
+  , padLeft (Pad 1) (str "]")
   , fillLine
   , txt (
       titleize (focusedViewName s) <> "-"
       <> titleize (focusedViewWidget s) <> " "
       )
   ]
 
+-- TODO
```
frasertweedale: TODO?

src/Storage/Notmuch.hs
romanofski: You know I think this is a remnant of me fiddling around. I'll have a look and change. Cheers!

src/UI/Actions.hs
romanofski: Will do!

src/UI/Actions.hs
romanofski: Aaaah I see. Thank you very much!

src/UI/Index/Keybindings.hs
romanofski: Yeah I simply took the default mutt behaviour, but I also realised that there is some value by keeping the cursor state. Also by keeping the list items toggled, after tagging is applied (Keybinding change needed too).

src/UI/Status/Main.hs
romanofski: Ah thanks for picking it up. I was trying to remember this morning to create a refactor issue, since all of this is currently a bit too rigid.

romanofski:
> @romanofski something busted in `unit` test suite:
> 
> ```
> test/TestActions.hs:32:18: error:
>     Module ‘UI.Utils’ does not export ‘selectedFiles’
>    |
> 32 | import UI.Utils (selectedFiles)
>    |                  ^^^^^^^^^^^^^
> cabal: Failed to build test:unit from purebred-0.1.0.0.
> ```
Hah! There is also an acceptance test failure I was investigating. Cheers for that. Saves me at least an additional run.

src/Storage/Notmuch.hs
romanofski: Hmm... so I looked into this and the problem is that `traverse` seems not to be the right abstraction. With `traverse` I get:

```
t [Notmuch.Message 0 mode]
```
as my `msgs`, while with `foldM` I get my list of messages returned. I've checked our notmuch binding which has:

```Haskell
class HasMessages a where
  messages
    :: (AsNotmuchError e, MonadError e m, MonadIO m)
    => a mode -> m [Message 0 mode]
```
In a way it does make sense to me (not sure if I get this right), since `traverse` is defined over `Applicative f`, while our monadic list would be one level deeper since `Monad` is a subclass of `Applicative`.

Yes?

romanofski:
Not done yet. Spend a few days to make the theme and colours more consistent, since it was becoming a bit of a hotchpotch with the additional toggled state (we now have to indicate: new mail, selected and toggled).

src/Storage/Notmuch.hs
romanofski: So I slept over this a night and I think I got it backwards. It should work since traverse already works for `Applicative` and since `Monad` is a subclass of `Applicative` it should also work for `Monad`... hrm

frasertweedale: @romanofski see https://github.com/purebred-mua/purebred/commit/9b0e18728b84bd25de7df01fcc21d2e6b0cf77e1.  For your consideration (i.e. it's not a showstopper; if you prefer `foldM` we'll keep it as-is).

romanofski: Hah no.. that looks good. I'll have to keep this in mind!

frasertweedale:
good to merge?

romanofski:
Yes :) Oof... took me quite a bit for this one.

frasertweedale:
It's great work mate.  Good on ya!

romanofski:
Thank you @frasertweedale :)

Issue #367, opened 2020-02-07
ci: add GHC 8.8 to matrix


frasertweedale:

romanofski:
This would involve editing the `.travis.yml` in the root directory of purebred to add GHC 8.8 to the matrix.

frasertweedale:
Oh, and 8.10 is out too :D

PR #382, opened 2020-04-08
ci: add 8.8 and 8.10 to matrix

frasertweedale:
Fixes: https://github.com/purebred-mua/purebred/issues/367

```
3babb53 (Fraser Tweedale, 4 hours ago)
   ci: cabal-install >= 3.0 does not know --symlink-bindir

   Fixes: https://github.com/purebred-mua/purebred/issues/367

2f4db4c (Fraser Tweedale, 5 hours ago)
   ci: add ghc 8.8 and 8.10 to matrix

5344f26 (Fraser Tweedale, 5 hours ago)
   ci: extract apt sources to common def

   We can use YAML anchor to define the APT sources only once. Unfortunately,
   YAML has no array merge or extend capability so we cannot do similar for
   the package list (because the versions of cabal and ghc vary).
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,18 @@ apt_sources: &apt_sources
 
 matrix:
   include:
-    - compiler: "ghc-8.6.3"
+    - compiler: "ghc-8.10.1"
+      language: c
+      env: NOTMUCHVER=0.28 GHCHEAD=true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -25,20 +25,33 @@ before_cache:
 
   - rm -rfv $HOME/.cabal/packages/head.hackage
 
+apt_sources: &apt_sources
```
romanofski, line 28: Nice!

Issue #379, opened 2020-03-11
Replying to an email duplicates the inline body


romanofski:
**Describe the bug**
When I reply to an email, the body is duplicated once saved.

**To Reproduce**
1. Open an email
2. Reply to this email which will show your editor.
3. Change body
4. Save it and exit, so you end up back in purebred
5. You should have two inline attachments: one is the original, but quoted mail, one is your changed reply in step 3

**Expected behavior**
A single inline attachment is shown.

**Screenshots**
See screen shot
![screenshot](https://user-images.githubusercontent.com/121047/76399383-3ca1c600-63ca-11ea-8411-53725d5806dc.png)

**Additional context**
This is a home made bug introduced in 535ebb66b0b0272b12a3f05fe32a9ff93486d256 by myself :)

romanofski:
I just noticed that this also happens, when resuming to compose a draft.

romanofski:
This relates to the pull request #361 and #294. Specifically the work @frasertweedale did regarding the `Action` refactoring.

PR #384, opened 2020-04-11
Use a more explicit way to update the attachment list

romanofski:
The way we updated the attachment list was by guessing the
operation. Well, if the list was empty the operation was pretty straight
forward: insert the item(s). However if the list was not empty the
guessing became harder: either replace or append. The ambiguity of this
resulted in this bug.

Instead of guessing, pass the operation we want to do after we've
invoked the operator to the function which invokes the editor. We know
exactly what we want to do with the list, after the editor has exited.

Fixes https://github.com/purebred-mua/purebred/issues/379

----

#

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -130,7 +130,7 @@ import Prelude hiding (readFile, unlines)
 import Data.Foldable (toList, traverse_)
 import Data.Functor.Identity (Identity(..))
 import Control.Lens
-       (_Just, to, at, ix, _1, _2, toListOf, traverse, traversed, has, snoc,
+       (_Just, to, at, ix, _1, _2, toListOf, traverse, traversed, has,
         filtered, set, over, preview, view, (&), firstOf, non, Traversal',
         Getting, Lens', folded, assign, modifying, preuse, use, uses
         , Ixed, Index, IxValue)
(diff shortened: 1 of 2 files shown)


src/UI/Actions.hs
```diff
@@ -1298,14 +1298,20 @@ makeAttachmentsFromSelected s = do
...
-    go parts l = foldr upsertPart l parts
     makeFullPath path = currentLine (view (asFileBrowser . fbSearchPath . E.editContentsL) s) </> path
 
+listAppendAttachments :: [MIMEMessage] -> L.List Name MIMEMessage -> L.List Name MIMEMessage
+listAppendAttachments parts l =
+  let append newPart l' =
+        l' &
+        over L.listElementsL (`snoc` newPart) .
+        set L.listSelectedL (Just (view (L.listElementsL . to length) l'))
+   in foldr append l parts
```
frasertweedale, line 1313: This seems to attach the new parts one by one.  Why not do a single `(<>)`?
```haskell
listAppendAttachments parts = listMoveTo (-1) . over L.listElementsL (<> parts)
```
Would that work?

romanofski: Ah of course! This is wonderful! Thank you so much.

Issue #383, opened 2020-04-10
"user can edit mail headers" intermittendly failing


romanofski:
**Describe the bug**
The user acceptance test "user can edit mail headers" seems to be intermittendly failing.

**To Reproduce**
Currently can not be 100% reproduced. It got picked up by our [cron
job](https://travis-ci.org/github/purebred-mua/purebred/jobs/673232694). Interestingly
a cron job before failed with the same job too:
https://travis-ci.org/github/purebred-mua/purebred/jobs/672707006 That
is the GHC 8.6.5 job.

**Expected behavior**
Job passes - always.

**Screenshots**
Log output:

```
user can edit mail headers:                               FAIL (1.92s)
    start composition           (0.03s)
    accept default              (0.03s)
    enter to: email             (0.03s)
    leave default               (0.03s)
    enter mail body
    exit insert mode in vim
    exit vim                    (0.03s)
    user can change from header (0.03s)
    append an email             (0.03s)
      src/Test/Tasty/Tmux.hs:445:
      Condition not met: 'Regex "Purebred:.*ComposeView-Attachments\\s+$^$"'.  Last capture:
      
                 From: "Joe Bloggs" <joe@foo.test>, testuser@foo.test                  
                  To: user@to.test                                                    
                  Cc:                                                                 
                 Bcc:                                                                 
             Subject:                                                                 
                                                                                      
       -- Attachments-----------------------------------------------------------------
       I --                                               text/plain; charset=us-ascii
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
                                                                                      
      Purebred: (0,46) [ New: 0  ]           ComposeView-Editor to Compose a new Mail 
      From: "Joe Bloggs" <joe@foo.test>, testuser@foo.test                            
      
      
[...]

PR #386, opened 2020-04-17
Change condition to assert the focus has changed

romanofski:
This test used an ambiguous condition to determine whether the editor
for a header field has closed and the focus switched back to the list of
attachments. Most times this didn't cause any problems. However
sometimes, this caused the test to fail with the editor still open,
since the condition was already met.

This swaps the two conditions around: first assert that the focus has
changed to the attachment list. Then assert that the field has the
additional text entered.

Fixes https://github.com/purebred-mua/purebred/issues/383


romanofski:
Lets see whether I'm lucky with this one or not ...

PR #392, opened 2020-05-01
Make the search path editor consistent with other editors

romanofski:
This adds additional key bindings to make the editor consistent with
other input widgets in Purebred.

Before using the search path editor was very confusing, since the only
way to move the focus back to the list (a.k.a file browser) was by
pressing Escape.

This patch changes the behaviour to allow the focus change to the list
with more key strokes.


romanofski:
One of those not-big-deal merge requests. Yet I didn't want to bother creating another issue for it or just leave it "broken", when I was able to simple just add those additional bindings and be done with it.

PR #393, opened 2020-05-05
Make hlint happy

romanofski:
Latest hlint (> 3.0) found 5 hints which caused our CI to fail. Adopt
the suggestions with this patch and make CI green again.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -111,6 +111,7 @@ data ListWithLength t a = ListWithLength (L.GenericList Name t a) (Maybe Int)
 
 listList :: Lens' (ListWithLength t a) (L.GenericList Name t a)
 listList f (ListWithLength a b) = (\a' -> ListWithLength a' b) <$> f a
+{-# ANN listList ("HLint: ignore Avoid lambda using `infix`" :: String) #-}
 
 listLength :: Lens' (ListWithLength t a) (Maybe Int)
 listLength f (ListWithLength a b) = (\b' -> ListWithLength a b') <$> f b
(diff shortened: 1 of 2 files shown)


src/Types.hs
```diff
@@ -110,7 +110,7 @@ data Name =
 data ListWithLength t a = ListWithLength (L.GenericList Name t a) (Maybe Int)
 
 listList :: Lens' (ListWithLength t a) (L.GenericList Name t a)
-listList f (ListWithLength a b) = (\a' -> ListWithLength a' b) <$> f a
```
frasertweedale, line 113: I'd prefer the Lens definitions be left as they were; infix notation is pretty awful here IMO.

So use a `{- ANN -}` module annotation to suppress this warning.  I'm sure there's already some modules where I did that.

frasertweedale: Ah it can be done per-function, e.g.: 

```haskell
listList = blah blah blah
{-# ANN listList ("HLint: ignore Avoid lambda" :: String) #-}
```

Likewise for the other functions.

src/UI/Actions.hs
```diff
@@ -1623,7 +1623,7 @@ pipeCommand' cmd
 editAttachment :: (MonadState AppState m, MonadIO m, MonadMask m) => m ()
 editAttachment = selectedItemHelper (asCompose . cAttachments) $ \m ->
   case preview (headers . contentDisposition . folded . dispositionType) m of
-    Just Inline -> invokeEditor' (\newPart l -> L.listModify (const newPart) l)
+    Just Inline -> invokeEditor' (L.listModify . const)
```
frasertweedale, line 1626: :+1:

src/Types.hs
romanofski: Ahh yep gotcha. Hopefully fixed now. I did this half-blind, since it seems the new hlint release also needs a higher GHC release than I currently have in my environment. Haven't had time to look into it.

romanofski:
Might drop the nix change to use hlint 3 or push it in a separate branch. Just want to see how this change goes now.

romanofski:
Alright good to go. Ready when you are :)

Issue #321, opened 2019-08-05
User Acceptance tests blocking forever in CI


romanofski:
**Describe the bug**
I've now run into this twice hence filing this bug. Travis errors out with:

```

travis_time:start:0245bd55
$ if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
Build profile: -w ghc-8.6.3 -O1
In order, the following will be built (use -v for more details):
 - purebred-0.1.0.0 (test:uat) (ephemeral targets)
 - purebred-0.1.0.0 (test:unit) (ephemeral targets)
Preprocessing test suite 'uat' for purebred-0.1.0.0..
Building test suite 'uat' for purebred-0.1.0.0..
Running 1 test suites...
Test suite uat: RUNNING...


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

**To Reproduce**
Unable to reproduce

**Expected behavior**
user acceptance tests run

**Additional context**
I currently don't have a theory.

frasertweedale:
Does it occur randomly or can it be reproduced reliably somehow?

romanofski:
Nah it's random and seems to be rare. When I first saw it I thought it might be the CI. I still think it's the CI, but since it happened a second time I'm not so sure any more.

romanofski:
Hm... we're using `cabal new-test` which, the last time I read the docs, doesn't automatically stream the results, but rather runs all tests and then shows the output. Could it be that an (output) buffer is full and therefore the entire thing hangs?

frasertweedale:
You might be onto something...

romanofski:
Happened here again: https://travis-ci.org/purebred-mua/purebred/jobs/614460305?utm_medium=notification&utm_source=github_status

romanofski:
Happened today again: https://travis-ci.org/purebred-mua/purebred/jobs/647519028?utm_medium=notification&utm_source=github_status

romanofski:
Happened again: https://travis-ci.org/github/purebred-mua/purebred/jobs/684070671

romanofski:
Interesting failure I found with the GHC-8.6.5 test: 
[log.txt](https://github.com/purebred-mua/purebred/files/4655723/log.txt)

[...]

PR #397, opened 2020-05-09
travis: Stream test progress

romanofski:
This streams the test output to the console in CI. This may be a
solution for #321 if the underlying cause is that the test run blocks
for whatever reason until it finishes.

If however the cause is something different, this patch may at least
shed some light into why CI kill's the job without output for 10min.

Relates to https://github.com/purebred-mua/purebred/issues/321

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,11 +42,11 @@ matrix:
       addons: {apt: {packages: [elinks,tmux,g++,libxapian-dev,libgmime-2.6-dev,libtalloc-dev,zlib1g-dev,ghc-ppa-tools,cabal-install-3.0,ghc-8.8.3], sources: *apt_sources}}
     - compiler: "ghc-8.6.5"
       language: c
-      env: NOTMUCHVER=0.28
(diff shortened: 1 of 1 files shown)


romanofski:
heh jeez... cabal.. why are the arguments so different :)

romanofski:
Updated commit message:

travis: Stream test progress

This streams the test output to the console in CI. This may be a
solution for #321 if the underlying cause is that no test output is
directed to stdout until the test command finishes.

If however the cause is something different, this patch may at least
shed some light into why CI kill's the job without output for 10min.

This patch comes with a small caveat tho. We default to the `new-test`
command. However, streaming support existed for older cabal versions
provided by the test command. Newer cabal versions (>= 3) provide the
streaming support with the `v2-test --test-show-details` command. This
is not supported with previous cabal-install releases. The old `test`
command in return does not understand the `--enable-tests` and
`--enable-benchmarks` options.

So to accommodate all, cabal-install for GHC 8.4.4 and 8.6.5 won't be
able to run any benchmarks in case they're enabled by environment
variable. Since we will probably drop those versions soon anyway, a
small sacrifice to make.

Relates to https://github.com/purebred-mua/purebred/issues/321

Issue #391, opened 2020-05-01
Filebrowser search path could end in a dead end


romanofski:
**Describe the bug**
Changing the search path in the file browser could have unforeseen consequences if we're using the editor. So far I was unable to open a gate to the fourth dimension. I'm still trying however.

**To Reproduce**
1. Compose an e-mail.
2. Open the file browser
3. Focus the search path editor (using `:`)
4. Replace the path with an invalid path and press `Escape`
5. Try to move up a parent with `CTRL+u`. The search path is now set to `.` and you won't be able to reset it to anything else.

**Expected behavior**
Allow to abort editing, so that the editor reset's it's input state to the previous input.

**Screenshots**
None

**Additional context**
I actually think that almost none of our input fields allow to discard invalid input on user request :/

romanofski:
With this one I'm giving the Brick filebrowser another attempt. Doesn't have the fancy unicode checkboxes, but I'm not really keen on maintaining our own instance of a file browser with potential problems.

PR #400, opened 2020-05-30
Handle garbage path input

romanofski:
This patch fixes a problem with the path editor setting the working
directory in the file browser. Since the file browser itself is not
checking whether the working directory is valid, it is up to the caller
to verify we're not setting anything invalid.

This patch checks whether the path for the working directory exists and
otherwise shows an error.

Fixes https://github.com/purebred-mua/purebred/issues/391


src/UI/Status/Main.hs
```diff
@@ -104,6 +105,9 @@ instance WithContext (ListWithLength t e) where
 instance WithContext (E.Editor Text Name) where
   renderContext _ = str . show . cursorPosition . view E.editContentsL
 
+instance WithContext (FB.FileBrowser Name) where
+  renderContext _ _ = str "filebrowser"
```
romanofski, line 109: Hm.. just realizing this little bit. The file browser widget itself has a status bar. Will have to see what makes sense to display in our status bar.

romanofski:
I just realised that there is an inconsistency with toggling. The Brick filebrowser default toggles items with using the `Space` key, while Purebred has bound it to `*`. The tricky thing comes from re-binding it to whatever you choose, since I've just noticed that with the current setup is tricky. I think if Brick can expose access to the underlying filebrowser widget's list we would be able to use what we already have in place for our `HasToggleableList` and everything would be peachy.

romanofski:
Went down the rabbit hole just to find out that it isn't as easy as I thought by exporting the list accessor on the file browser widget. Why? Because Brick's file browser is storing the toggled items in a separate `Set`, while we keep the state on the list item itself.
I've added an additional `Action` to toggle file browser items. It's a bit meh, but should suffice for now.

frasertweedale:
I'm ok with it if you want to merge.  More LOC removed than added :+1: .  I don't see any showstoppers

Issue #395, opened 2020-05-06
MessageNotFound after new mail was sent


romanofski:
**Describe the bug**
When I sent an email, the view goes back to the list of threads. When I try to re-open the same message, I get a MessageNotFound error. For example:

MessageNotFound "notmuch-sha1-37d366f6ef51d6b94836805748103ec126d391c8"

**To Reproduce**
1. (Perhaps a custom query needs to be executed first)
2. Compose a new mail from an old mail by opening a thread and pressing `e`.
3. Sent the mail.
4. Back on the thread view try to open the thread again.

Custom query first:
1. Use a custom query to find a few threads: e.g. to:<receipient>
2. Start the composition from there

**Expected behavior**
Not a `MessageNotFound` error.

**Screenshots**
None

**Additional context**
Typically I reload the thread list after going back. Perhaps the keybindings are configured like that too and should be investigated.

romanofski:
Just checked with composing a new mail from the default query after start up, and the problem doesn't exist once the mail is sent and Purebred went back to the threads view.

romanofski:
I think it has something to do with the mail being sent, falling into the result list of the search the threadview is set to. For example if the search is `to:john` and we're sending a new mail to `john`, the thread list is now obsolete, since the just sent mail isn't included ...

romanofski:
This one is nasty. The source of the problem is the `composeAsNew` action, which removes the file path from the notmuch index. I guess that's what we want when dealing with draft mails, since the drafts are getting removed and turned into mails we want to sent. However, not really something we want to do for existing mails, like mails we've sent.

So in this instance, what happens is, that the mail is removed from the index. The mail is composed as new, changed, sent and re-indexed. The view returns to the original query and obviously the original mail we created a new compose from doesn't exist any more.

frasertweedale:
@romanofski can we handle it in a similar way to how we update tags on the `NotmuchMail` after editing.  But instead of just modifying the selected `NotmuchMail` on the thread view, we replace it with a new one?

frasertweedale:
Hm I see the problem... OK here's another idea.  The message "key" in libnotmuch is the MessageID.  If we write (and index) draft mail with message ID, and keep that message ID the same for `composeAsNew`, then indexing and removal can just remove *files* from the libnotmuch message object, rather than removing the message from the notmuch db then adding a new one.  Then the same message object previously referred to shall still exist.

Maybe I am misunderstanding the scenario... but let me know what you think.

[...]

Issue #351, opened 2019-12-02
Don't delete Drafts mail before mail is sent


romanofski:
**Is your feature request related to a problem? Please describe.**
When editing an e-mail we store the temporary mail in the "Drafts" folder in order for an external editor to edit it. Once that is done control is handed back to purebred and the temporary file in the Drafts folder is removed. The mail is such is sitting in it's serialised form as data in purebred.

However if purebred crashes here, the mail is lost which is not very good. The status quo is due to the fact that we currently don't have a good way to find the draft mail again in order to update it (e.g. if the user decides to re-edit it again).

**Describe the solution you'd like**
Store any draft (first or updated) in the "Drafts" folder until the mail has been successfully sent. I think I can live if it doesn't reflect the latest changes to the `To:` field or `Subject:` line, but the mail body is important to not get lost.

**Describe alternatives you've considered**
None

**Additional context**
I realised it when adding a test for #350 because the acceptance test asserts that the "Draft" mail is empty after editing. That is a state I don't want Purebred to be in if it wants to become a productive tool.

PR #402, opened 2020-06-23
Workaround for composeAsNew

romanofski:
Initially composeAsNew was implemented with the idea to continue draft
mails. The mails are stored in the file system and indexed in
notmuch. Once the draft is picked up again, the mail is unindexed and
removed from the file system until it's sent.

However, if we use the same function on existing mails in order to
use the mail as a template for a new mail, we actually unindex and
remove the mail mistakenly from notmuch and the file system. Instead all
we want is a copy for composing.

This is a workaround for now, since ideally we want to keep all
compositions as a draft on the file system (see #351).

This patch only unindexes/removes draft mails, any other mail contents
are simply copied for composition.

Fixes https://github.com/purebred-mua/purebred/issues/395

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -1225,11 +1225,25 @@ composeAsNew = Action ["edit mail as new"] $
       (assignError (GenericError "No mail selected"))
       (\mail -> do
         pmail <- use (asMailView . mvMail)
-        dbpath <- use (asConfig . confNotmuch . nmDatabase)
         charsets <- use (asConfig . confCharsets)
-        runExceptT (Notmuch.mailFilepath mail dbpath >>= Notmuch.unindexFilePath dbpath)
+        runExceptT (specialCaseForDraft mail)
           >>= either assignError (const $ assign asCompose (newComposeFromMail charsets pmail))
       )
+
+-- | This is a special case workaround for draft mails used by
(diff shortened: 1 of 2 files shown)


romanofski:
I took the cheap way out for now, leaving it up to #351 to deal with it.

src/UI/Actions.hs
```diff
@@ -1225,12 +1225,27 @@ composeAsNew = Action ["edit mail as new"] $
...
+-- result in creating a copy mail object for non-draft mails.
+--
+specialCaseForDraft ::
+     (MonadState AppState m, MonadIO m, MonadError Error m)
+  => NotmuchMail
+  -> m ()
+specialCaseForDraft mail = do
+  dbpath <- use (asConfig . confNotmuch . nmDatabase)
+  draftTag <- use (asConfig . confNotmuch . nmDraftTag)
+  if draftTag `elem` view mailTags mail
```
frasertweedale, line 1245: this could be:
```haskell
when draftTag `elem` view mailTags mail $
  Notmuch.mailFilepath mail dbpath >>= Notmuch.unindexFilePath dbpath
```

romanofski: Ah cheers! Yes of course! Much better :)

Issue #336, opened 2019-10-22
Error message can mess up the terminal


romanofski:
**Describe the bug**
If the editor process hangs and is killed, the displayed error message can mess up the terminal (I presume).

**To Reproduce**
1. Start editing an email until you're dropped into your editor (e.g. vim)
2. Externally kill the `vim` process.
3. The compose editor now displays the error message.
4. It should be difficult to answer the confirmation dialog in order keep/discard the draft (`Enter` doesn't seem to work)

**Expected behavior**
Displaying the error message is fine (even though it's a wall of text), but the error message needs the same treatment we've introduced with any form of output to avoid messing with the terminal. See `Purebred.Types.IFC`.

**Additional context**
I'm not sure if the error message really is the culprit, but I thought I'd better file it in order to look into this issue.

romanofski:
I looked into this briefly today and I have my doubts whether this is caused showing the error message or output from the process. I removed showing anything at all (other than, well there was an error) and still had trouble with the key input. Also after quitting purebred, the terminal was b0rked until a `reset` too.

romanofski:
Made some progress here. The terminal settings after I kill vim, look like this:
```
$  stty
speed 38400 baud; line = 0; min = 1; time = 0;
-icrnl -onlcr -isig -icanon -iexten -echo -echoe
```
which is resulting in this weird terminal behaviour. In comparison my normal terminal settings after I ran a `reset`:

```
$ stty
speed 38400 baud; line = 0;
-brkint -imaxbel
```

I'll have to dig when where vim set's the terminal back to it's original state. I presume if the editor dies, that clean up isn't happening.

romanofski:
Diving deeper, I checked out the code of other terminal based mail clients and first thought starting the editor through a sub-shell must be it. I used a small reproducer and can rule out that the sub-shell would do any terminal setting/clean ups.

That left me with the framework to do any terminal work. I dived into [https://github.com/urwid/urwid](urwid)s source code and found that urwid is handling the terminal in cases of stopping/starting (see https://github.com/urwid/urwid/blob/master/urwid/display_common.py#L735) . I'll check out brick/vty whether it can handle term changes and prepare a patch in case it doesn't (or it doesn't work properly).

jtdaugherty:
[...]

PR #403, opened 2020-07-18
Refactor focusing of views and widgets

romanofski:
Purebred had two chain combinators:

a) `chain` for chaining Action's of the same view/widget
b) `chain'` for chaining Action's of *different* view/widget

Furthermore, we had an action called `focus` which actually didn't focus
a view/widget, but rather ran setup/teardown code when view and widgets
were changed.

This is confusing.

This patch hopes to improve how Purebred makes focus changes.

a) The `chain'` combinator becomes 'focus' in order to make clear, that
we're focusing a new widget in a potentially different view.

b) The former `focus` Action is removed instead.

c) The `switchFocus` class method has been renamed to `onFocusSwitch` to
indicate the fact that this is not actually changing the focus to a view
or widget. Rather it runs code during a switch of focus to a different
widget.

d) The explicit view/widget specifier for Actions which can be used on
multiple views/widgets is now made on the right hand side of the `focus`
combinator.

Examples:

A former keybinding like:

    Keybinding (V.EvKey (V.KChar 'g') [V.MCtrl]) (abort `chain'` focus @'ComposeView @'ComposeListOfAttachments `chain` continue)

becomes:

     Keybinding (V.EvKey (V.KChar 'g') [V.MCtrl]) (abort `focus` continue @'ComposeView @'ComposeListOfAttachments)

This former keybinding:

     Keybinding (V.EvKey V.KEnter []) (handleConfirm `chain'` focus @'Threads @'ListOfThreads `chain'` reloadList `chain` continue)

becomes:

     Keybinding (V.EvKey V.KEnter []) (handleConfirm `focus` reloadList `chain` continue)

----

#

### DIFF
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -28,15 +28,15 @@ import Purebred
 
 scrollKeybindings :: ('Scrollable' w) => ['Keybinding' v w]
 scrollKeybindings =
-  [ 'Keybinding' (EvKey (KChar 'j') []) ('scrollDown' ``chain`` 'continue')
-  , Keybinding (EvKey (KChar 'k') []) ('scrollUp' \`chain\` continue)
-  , Keybinding (EvKey (KChar 'd') []) ('scrollPageDown' \`chain\` continue)
-  , Keybinding (EvKey (KChar 'u') []) ('scrollPageUp' \`chain\` continue)
+  [ 'Keybinding' (EvKey (KChar 'j') []) ('scrollDown' \`chain\` 'continue')
+  , Keybinding (EvKey (KChar 'k') []) ('scrollUp' \`chain\` 'continue')
+  , Keybinding (EvKey (KChar 'd') []) ('scrollPageDown' \`chain\` 'continue')
+  , Keybinding (EvKey (KChar 'u') []) ('scrollPageUp' \`chain\` 'continue')
   ]
 
 mailViewKeybindings =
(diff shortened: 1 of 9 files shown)


romanofski:
Would make my change for #336 easier, since I want Purebred to abort composition and go back to the thread view in case the editor dies.

src/Purebred.hs
```diff
@@ -28,15 +28,15 @@ import Purebred
...
 scrollKeybindings :: ('Scrollable' w) => ['Keybinding' v w]
 scrollKeybindings =
-  [ 'Keybinding' (EvKey (KChar 'j') []) ('scrollDown' ``chain`` 'continue')
-  , Keybinding (EvKey (KChar 'k') []) ('scrollUp' \`chain\` continue)
-  , Keybinding (EvKey (KChar 'd') []) ('scrollPageDown' \`chain\` continue)
-  , Keybinding (EvKey (KChar 'u') []) ('scrollPageUp' \`chain\` continue)
+  [ 'Keybinding' (EvKey (KChar 'j') []) ('scrollDown' \`chain\` 'continue')
+  , Keybinding (EvKey (KChar 'k') []) ('scrollUp' \`chain\` `continue`)
+  , Keybinding (EvKey (KChar 'd') []) ('scrollPageDown' \`chain\` `continue`)
+  , Keybinding (EvKey (KChar 'u') []) ('scrollPageUp' \`chain\` `continue`)
```
frasertweedale, line 34: Are backticks around `continue` correct?

romanofski: Hmm you're right. Thanks for spotting!

PR #404, opened 2020-08-01
Fall back to previous view in case of an editor error

romanofski:
```
30d9d60 Bump purebred-email to latest 0.4.2 release

Apart from the fixes, we also need the higher package boundaries,
otherwise the nix builds will fail.

eb59fd9 nix: use new release and GHC 8.8.4

This pin's to a new nix release to avoid CI mostly fetching cached
packages instead of rebuilding the (Nix) world.

With bumping minimum versions for vty and brick, the minimum boundaries
of all transitive packages has shifted upwards, causing a lot of
derivations to be rebuilt due to the fact that they are not included in
the package closure of the former pinned versions.

a3b32ce Fall back to previous view in case of editor error

When Purebred is suspended and we hand control over to our editor, it is
possible that the editor errors (crashes, doesn't start, wrong editor,
yadda yadda).

Currently Purebred would simply go to the compose editor with an empty
attachment. That means that we could possibly send an email with an
empty body.

Instead, parameterize the `invokeEditor` function in which we give a
fallback view and widget. In cases of an error, we fall back to this
view and the focused widget.

Fixes: https://github.com/purebred-mua/purebred/issues/336
```

### DIFF
--- a/.nix/brick.nix
+++ b/.nix/brick.nix
@@ -6,8 +6,8 @@
 }:
 mkDerivation {
   pname = "brick";
-  version = "0.54";
-  sha256 = "1bce8104859986a3f871751de34f1876a3ece7632fe981d0cb267a1e00074281";
+  version = "0.55";
+  sha256 = "de09e03e8223ed10d0bddcbb836726b5849b7f9a092ab7b05942952311dca158";
   isLibrary = true;
   isExecutable = true;
   libraryHaskellDepends = [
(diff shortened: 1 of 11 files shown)


romanofski:
Slipped in another patch. Behaviour was bothering me for quite a while when opening HTML attachments with an external browser.

frasertweedale:
Yee haw!

Issue #327, opened 2019-09-14
Intermittent test failure in file browser test


romanofski:
**Describe the bug**
I've had a test run which seemed to have failed with an intermittent failure, since it only happend on the nix build. The last screenshot seems to suggest that for whatever reason the list item wasn't selecting the last remaining entity to trigger the error message.

Perhaps it is again a too generic intermittent step which causes the tests to skip too quickly?

**To Reproduce**
No idea at this point.

**Expected behavior**
Test run without a failure

**Screenshots**
See travis build [log.txt](https://github.com/purebred-mua/purebred/files/3612591/log.txt)

**Additional context**
None

romanofski:
Happened today again:

![Screenshot_2019-10-22 Job #1171 2 - purebred-mua purebred - Travis CI](https://user-images.githubusercontent.com/121047/67271284-937f2f80-f4fd-11e9-99bf-cc3a2e010383.png)

romanofski:
Happened again here: https://travis-ci.org/purebred-mua/purebred/jobs/606239891?utm_medium=notification&utm_source=email

romanofski:
I've looked into this a little bit today. The test steps seem fine and I can't find any sub strings which are too generic so the test run would skip ahead.

The only hypothesis I could come up with is perhaps the attachment list and it's length indicator in the status bar. Our test framework doesn't really let us assert that a string has been removed from the previous screen (maybe an feature we should implement?). Instead of asserting that the last file was deleted, because it's now missing, we use the item indicator in the status bar and assert with the list length, e.g. `Item 1 of 2`.

Now I'm wondering whether perhaps the list length is recalculated quicker (since it's asynchronous) than the redrawing of the entire UI can take place. In the screen shot the deleted list entry still looks like it's selected, even tho the entry is gone.

If that's a theory to pursue, I'm currently not sure what else to assert on which would be a stable indicator that the file is *not* in the list/UI.

@frasertweedale does that sound plausible as a possible cause?

romanofski:
Happened again today: https://travis-ci.org/purebred-mua/purebred/jobs/613420288?utm_medium=notification&utm_source=github_status

romanofski:
Had this come up again today in a test failure: https://travis-ci.org/purebred-mua/purebred/jobs/629258818?utm_medium=notification&utm_source=email

Unfortunately it seems, this issue isn't fixed yet :/

romanofski:
[...]

PR #405, opened 2020-08-02
Fix intermittendly failing file browser test

romanofski:
The change 47c72c27a93a0098749bf970c5ac2fe4482750e2 tried to address the
intermittendly failing test, but I think made the mistake of
"hardcoding" the file we assume to be removing from the attachment list.

The problem however is, that the list of files we're adding as
attachment varies with the environment. While the `screenshot.png` is
probably the attachment in most cases, it isn't the attachment
in **all** cases. The assertion that the next screen does not show a
file which we never added is always true.

Instead, firstly, this patch asserts that the last filename is actually
listed as an attachment. Secondly, we assert that when the attachment is
removed the last file from before is gone. Furthermore we also check
whether the item count has changed.

Fixes: https://github.com/purebred-mua/purebred/issues/327

### DIFF
--- a/test/TestUserAcceptance.hs
+++ b/test/TestUserAcceptance.hs
@@ -144,7 +144,8 @@ testReloadsThreadListAfterReply = purebredTmuxSession "reloads list of threads" 
     sendKeys "e" (Substring "Testmail with whitespace in the subject")
 
     step "navigate to latest attachment"
-    sendKeys "Down" (Substring "Item 2 of 2")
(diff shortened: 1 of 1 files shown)


frasertweedale:
Wow, will this really fix it forever?  merge when green, I don't have the headspace to work this all out rn but I trust you mate!

romanofski:
Cheers! Yeah it annoyed me quite a bit for my last pull request. I hope I got now all bases covered.

PR #409, opened 2021-01-07
Move to Github Actions

romanofski:
Travis is going away for Open Source projects so this patch moves our
project to Github Actions.


romanofski:
Don't merge yet.. I want to grab the two or three additional jobs we had added ...

frasertweedale:
@romanofski you might be able to use the haskell-ci `--github-patches=FILE` option to apply a patch that adds our secret herbs and spices.

PR #413, opened 2021-01-15
Better notification system

romanofski:
```
commit c248c50ce861d60f21f6b278145a1f2cbf749b4b (HEAD -> feature/286-notification-system, origin/feature/286-notification-system)
Date:   Thu Jan 14 06:10:08 2021 +1000

    Remove unused imports

commit e2e70d8c3d4cb9a6e15e811c419415458deba3b7
Date:   Tue Jan 12 19:37:44 2021 +1000

    Use the widget name to avoid overflowing notifications
    
    Some editors use real time validation. That is, per character we fire of
    a thread to validate, but kill the thread if the next character is
    received in less than 500ms.
    
    However, when the editor widget gets input too fast (e.g. tmux driven
    tests or to a lesser extend a copy & paste[1]) and the last character
    finishes the input (e.g. `Enter` to switch to the next input) the
    validation message "overflows" into the next widget.
    
    Instead, use the context information on the user message, to only show
    user messages for widgets still visible in the UI.
    
    [1] There is specific copy & paste support (aka bracketed paste). Only
    the quick succession of receiving input is meant in this context.
    
    Fixes: https://github.com/purebred-mua/purebred/issues/412

commit 52fcb19eab1db11777baf2349072ff53b8cd4edf
Date:   Tue Jan 12 19:36:47 2021 +1000

    Fix validation
    
    The problem here is that `$>` sequenced the two actions so that
    `schedule` was the remaining action to be evaluated. That caused a
    scheduling of a thread per character input without killing any of the
    remaining threads.

commit 1ac5c3b6bdbf7952898bfa43182f37f703bedb04
Date:   Thu Aug 13 07:08:47 2020 +1000

    Better notification system
    
    This implements an improvement to the current notification system.
    
    The `UserMessage` differenciates three levels of severity:
    
    * Info
    * Warning
    * Error
    
    That helps to remove most of the `GenericError` messages which were
    misused before as basic feedback.
    
    Furthermore this adds a context, so we can later filter out if messages
    are relevant to be visible or not.
    
    Fixes: https://github.com/purebred-mua/purebred/issues/286

```

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -1,5 +1,5 @@
 -- This file is part of purebred
--- Copyright (C) 2017-2019 Róman Joost and Fraser Tweedale
+-- Copyright (C) 2017-2021 Róman Joost and Fraser Tweedale
 --
 -- purebred is free software: you can redistribute it and/or modify
 -- it under the terms of the GNU Affero General Public License as published by
@@ -32,7 +32,7 @@ module Types
   , asThreadsView
   , asMailView
   , asCompose
-  , asError
+  , asUserMessage
   , asViews
   , asFileBrowser
   , asLocalTime
@@ -255,6 +255,12 @@ module Types
   , rsFree
   , decodeLenient
   , Generation(..)
+
+  -- ** User notifications
+  , UserMessage(..)
+  , umContext
+  , umSeverity
+  , MessageSeverity(..)
   ) where
 
 import Prelude hiding (Word)
@@ -926,7 +932,7 @@ data AppState = AppState
     , _asThreadsView :: ThreadsView
     , _asMailView  :: MailView
     , _asCompose   :: Compose  -- ^ state to keep when user creates a new mail
-    , _asError     :: Maybe Error -- ^ in case of errors, show this error message
+    , _asUserMessage :: Maybe UserMessage
     , _asViews     :: ViewSettings -- ^ stores widget and focus information
     , _asFileBrowser :: FileBrowser
     , _asLocalTime :: UTCTime
@@ -945,11 +951,11 @@ asMailView = lens _asMailView (\appstate x -> appstate { _asMailView = x })
 asCompose :: Lens' AppState Compose
 asCompose = lens _asCompose (\appstate x -> appstate { _asCompose = x })
 
-asError :: Lens' AppState (Maybe Error)
-asError = lens _asError (\appstate x -> appstate { _asError = x })
+asUserMessage :: Lens' AppState (Maybe UserMessage)
+asUserMessage = lens _asUserMessage (\appstate x -> appstate { _asUserMessage = x })
 
 asViews :: Lens' AppState ViewSettings
-asViews f (AppState a b c d e g h i j) = fmap (\g' -> AppState a b c d e g' h i j) (f g)
+asViews = lens _asViews (\appstate x -> appstate { _asViews = x })
 
 asFileBrowser :: Lens' AppState FileBrowser
 asFileBrowser = lens _asFileBrowser (\as x -> as { _asFileBrowser = x })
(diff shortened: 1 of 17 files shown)


src/UI/App.hs
```diff
@@ -155,7 +155,16 @@ appEvent s (T.AppEvent ev) = case ev of
     then set (asThreadsView . miThreads . listLength) (Just n) s
     else s
   NotifyNewMailArrived n -> M.continue (set (asThreadsView . miNewMail) n s)
-  InputValidated l err -> M.continue (s & set l err . set (asAsync . aValidation) Nothing)
+  InputValidated l err ->
```
romanofski, line 158: Wondering if this can be done in a ... well less convoluted way. But currently the best I can come up with.

src/UI/App.hs
```diff
@@ -1,9 +1,9 @@
 -- This file is part of purebred
--- Copyright (C) 2017-2019 Róman Joost and Fraser Tweedale
+-- Copyright (C) 2017-2021 Róman Joost and Fraser Tweedale
 --
 -- purebred is free software: you can redistribute it and/or modify
 -- it under the terms of the GNU Affero General Public License as published by
--- the Free Software Foundation, either version 3 of the License, or
+-- the Free Software Foundation, either versior 3 of the License, or
```
frasertweedale, line 6: typo

romanofski: Heh - when did this one snuck in...

src/UI/App.hs
frasertweedale: It seems OK to me.  Only suggestions:
- redundant parens in `(Just msg)` pattern
- You could lift the `M.continue` or even `(M.continue . ($ s))` outside the `case` expression, as those are repeated in every branch.
- A bit more commentary about what each case *means* would be nice :)

src/Types.hs
```diff
@@ -1188,5 +1194,25 @@ data PurebredEvent
   = NotifyNumThreads Int
                      Generation
   | NotifyNewMailArrived Int
-  | InputValidated (Lens' AppState (Maybe Error))
-                   (Maybe Error)
+  | InputValidated (Lens' AppState (Maybe UserMessage))
+                   (Maybe UserMessage)
```
frasertweedale, line 1198: Could you please add comments describing these fields?  At a glance, without reading other code, it seems like this event includes both the message, and a lens to the part of the `AppState` to which the message relates.  And the handler would be responsible for setting it into the `AppState`.  I am not sure if the intuition is correct though.  So we should describe it in more detail.

src/UI/Notifications.hs
```diff
@@ -0,0 +1,81 @@
...
+
+
+makeInfo :: Name -> T.Text -> UserMessage
+makeInfo n = UserMessage n . Info
+
+makeWarning :: Name -> T.Text -> UserMessage
+makeWarning n = UserMessage n . Warning
+
+makeError :: Name -> Error -> UserMessage
+makeError n = UserMessage n . Error
```
frasertweedale, line 81: The functions dealing with `Error` are a bit "odd one out", with the others dealing with `Text`.  Not a criticism really, just an observation.   It might be nice to have a shortcut function such as:
```haskell
makeGenericError :: Name -> T.Text -> UserMessage
makeGenericError n = UserMessage n . Error . GenericError
```

frasertweedale: Probably time to change the argument of `GenericError` constructor from `String` to `Text` too.

frasertweedale:
@romanofski a bunch of general comments.  No showstoppers.  I'll give it a test drive in the next couple days.

src/UI/Notifications.hs
romanofski: Hm.. see I think we should throw out the `GenericError`.. I've hopefully changed all the `GenericError`s to warnings and IMHO either we have a specific error to show or we use a different severity. But I'd say there isn't really something like a "generic" error...

src/UI/App.hs
romanofski: Hm.. interesting that hlint didn't pick up on the redundant parenthesis. Everything else should be done. Cheers!

src/Types.hs
romanofski: Ok. Done.

src/UI/Notifications.hs
romanofski: Yeah I guess the `Error` still sticks out as odd, but I hope I made it less weird. Also removed the `GenericError`. Either we use a special `Error` constructor or it's a different severity.

frasertweedale:
@romanofski sorry I still didn't review this yet.  If you want to just merge it, go ahead :)

romanofski:
That sounds fine with me. I'll merge ;)

PR #415, opened 2021-02-03
add plugin system

### LINKED ISSUE #414: Add user-agent to outgoing mail
**Describe the solution you'd like**

It would be nice to include `User-Agent` header in outgoing mail, containing purebred version, and versions of any important components (purebred-email, hs-notmuch, libnotmuch).  In particular this could help with bug reporting.  And it is also nice to let the world know that you are using Purebred :)

**Describe alternatives you've considered**

Status quo (no user-agent header).  It is not mandatory.

**Additional context**

Although `User-Agent` for mail is not formally defined by IETF, it is defined for netnews (https://tools.ietf.org/html/rfc5536#section-3.2.13) and we should probably conform to that specification.

frasertweedale:
Add a plugin system.  The design of the system is described in [1]
Also provide a built-in plugin to add the User-Agent header to
outgoing mail (fixes #414).

Further improvements to the system and documentation will follow in
subsequent commits.

[1] https://frasertweedale.github.io/blog-fp/posts/2021-02-02-plugin-system-prototype.html

Fixes: https://github.com/purebred-mua/purebred/issues/414

### DIFF
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -172,7 +172,7 @@ import Types
 import Error
 import Purebred.Plugin
 import Purebred.Plugin.Internal
-  ( configHook )
+  ( configHook, pluginBuiltIn, pluginName )
 
 -- re-exports for configuration
 import qualified Graphics.Vty
@@ -183,7 +183,7 @@ import Brick.Main (customMain)
 import Brick.Types (Next)
 import Brick.Util (on, fg, bg)
 import Brick.AttrMap (AttrName, applyAttrMappings)
-import Control.Lens ((&), _head, over, preview, set, view)
+import Control.Lens ((&), _head, over, preview, set, view, views)
 import Data.Text.Lens (packed)
 import Data.MIME (Mailbox(..), AddrSpec(..), Domain(..))
 
@@ -193,8 +193,8 @@ data AppConfig = AppConfig
     , debugFile :: Maybe FilePath
     }
 
-appconfig :: Parser AppConfig
-appconfig = AppConfig
+appconfig :: String -> Parser AppConfig
+appconfig verInfo = AppConfig
   <$> optional
     ( Builder.option Builder.str
       ( long "database"
@@ -216,7 +216,7 @@ appconfig = AppConfig
       <> help "Write debug information to FILE"
       )
     )
-  <* Builder.infoOption versionString
+  <* Builder.infoOption verInfo
     ( long "version"
     <> short 'v'
     <> help "Print the Purebred version and exit"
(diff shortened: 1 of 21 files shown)


frasertweedale:
Ooof, this fails in Haddock with a compile error related to the code itself.  i.e. a bug in haddock, probably?

frasertweedale:
Hm, it's caused by a difference in haddock's use of `.hs-boot` files. Probably still a bug, but I've worked around it.

romanofski:
Hey @frasertweedale, this is awesome! I'd merge it right away, however I think we should clarify how this plugin system sits with what we currently use with https://github.com/purebred-mua/purebred-icu. Will `purebred-icu` remain a different plugin or would we be migrating it? I guess what I'm getting at, perhaps it makes sense to have different points of Purebred and it's config pluggable, or rather have one spot, where plugins can be loaded?

frasertweedale:
Nono, do not merge it yet.  Lots more goodness to come.

purebred-icu is a *kind* of plugin, but doesn't hook anything, just twiddles some config.  However, it could be integrated with the plugin system by defining an "initConfig" hook that gets access to twiddle the config.  Which is a reasonable way to do it, and would provide consistent UX.  So I will probably migrate it.

frasertweedale:
@romanofski ready for review now.  I did implement a ConfigHook.  I would like to rename `purebredWithPlugins` to `purebred`, i.e. make it the primary (or only) entry point - a backwards incompatible change.  But LMK what you think about that.  It can be done later as a small and separate change, anyhow.

I look forward to adding more hooks to do interesting things :)  PGP plugin is the next yak to shave, while I avoid the dreaded crash bug :S

src/Purebred.hs
```diff
@@ -220,19 +225,42 @@ appconfig = AppConfig
 versionString :: String
 versionString = showVersion version
 
-optParser :: ParserInfo AppConfig
-optParser = info
-  (appconfig <**> helper)
+-- | Print full version info about plugins
+fullVersionInfo :: [PluginDict] -> String
```
romanofski, line 229: Ah nice!

src/Purebred/Plugin.hs
```diff
@@ -0,0 +1,161 @@
+-- This file is part of purebred
```
romanofski, line 1: Good stuff! Especially all the documentation!

romanofski (approved):

romanofski:
> @romanofski ready for review now. I did implement a ConfigHook. I would like to rename `purebredWithPlugins` to `purebred`, i.e. make it the primary (or only) entry point - a backwards incompatible change. But LMK what you think about that. It can be done later as a small and separate change, anyhow.

Yeah go ahead. It's not that we've released 1.0 or any version as of yet. So IMHO totally fine to make a backwards incompatible change.

PR #417, opened 2021-02-10
fix crash in substring search

frasertweedale:
Purebred crashes when you enter an empty substring search term, due
to non-total Data.Text.breakOnAll.  Treat empty search term as a
"reset" operation and add UAT regression test.


test/TestUserAcceptance.hs
```diff
@@ -492,6 +492,12 @@ testSubstringMatchesAreCleared = purebredTmuxSession "substring match indicator
     step "enter needle and show results"
     sendKeys "et\r" (Substring "1 of 20 matches")
 
+    step "begin substring search"
```
romanofski, line 495: Nice!

romanofski (approved):

PR #424, opened 2021-03-06
Solarized Dark theme adjustments

romanofski:
These all small tweaks to address a few small glitches:

* Selected list items had little contrast and therefore were hard to read.
* Toggled list items also had little contrast.
* The background of the editor was explicitly set to black. Simply use
  the background of the terminal which in a (solarized) dark setting
  should be black.


romanofski:
Before: 
before
After:
after

Now it's easier to read the German spam mails ;)

Issue #419, opened 2021-02-18
Group reply


romanofski:
**Is your feature request related to a problem? Please describe.**
Purebred is missing replying to an entire group.

**Describe the solution you'd like**
I'd like a way to reply to a group. Either by keystroke (e.g. `g` or by other means). Ideally, it would just be a some form of tab cycling in the composition view, where I can choose to reply to the entire group instead of the individual.

**Describe alternatives you've considered**
None.

**Additional context**
None.

frasertweedale:
I think initial goal of separate actions for single-reply and reply-all (and maybe also list-reply) is fine.  Just like mutt.

Altering an in-progress composition seems much more complicated: look up message id of in-reply-to field; extract recipients; modify current composition.  I think we can definitely kick that can down the road, possibly forever.  But if you feel so strongly you want it... who am I to stand in your way?

frasertweedale:
I put it in milestone 1.0.  Reply-all is an essential feature.

romanofski:
Yeah I actually think altering an in-progress composition should be trivial (I think). The composition is held in memory anyway at this point in time (maybe shouldn't). So all it would take is probably a better data type for the address and a focus ring, which should allow us to cycle between the typical `reply`, `reply all` and don't know other options.

... unless? I've missed something.

frasertweedale:
The current composition doesn't hold the actual mail being replied to though, does it?  It will have to look that up to find out who were the recipients of the email being replied to.  But maybe it is I who is missing something...

romanofski:
No it doesn't, that's correct. However we pre-populate the UI fields with items we parse out of the mail we're replying to. That's where I would think I could grab all necessary information for the options: either the single-reply as it is, all addresses for a possible group reply and.. well whatever possibility there is (e.g. mailing lists). In the composition view, you could repeatedly press a key to then cycle through the options to set the `To:` field. When we write out the `ByteString`, we set whatever has been set in the UI fields.

frasertweedale:
On Fri, Feb 19, 2021 at 12:32:59AM -0800, Roman Joost wrote:

> No it doesn't, that's correct. However we pre-populate the UI
> fields with items we parse out of the mail we're replying to.
> That's where I would think I could grab all necessary information
> for the options: either the single-reply as it is, all addresses
> for a possible group reply and.. well whatever possibility there
> is (e.g. mailing lists). In the composition view, you could
> repeatedly press a key to then cycle through the options to set
[...]

PR #425, opened 2021-03-11
group reply

frasertweedale:
Implement group reply (#419).  Requires
https://github.com/purebred-mua/purebred-email/pull/63.

Also fix a bug where non-ASCII headers were presented with raw encoded-words
when replying.  This affected UI only, but was quite annoying.

```
88842ca (Fraser Tweedale, 2 days ago)
   add group reply

   Implement the group reply action and bind it to 'R'.

   Fixes: https://github.com/purebred-mua/purebred/issues/419

ed56ed5 (Fraser Tweedale, 3 days ago)
   reply: render non-ASCII headers properly

   When replying, purebred does not perform encoded-word decoding of the
   values of the To/From/Subject header fields.  As a consequence, raw
   encoded-words can appear in these values.

   Use the specific header optics provided by purebred-email and remove our
   buggy and unneeded helper functions.  As a result, headers with 
   encoded-words are presented with the values decoded.

   Also add a regression test to UAT suite.

55ae926 (Fraser Tweedale, 7 weeks ago)
   use purebred-email 0.5

   Bump to purebred-email 0.5.  Note the following:

   - Outbound mail now uses zoned time in the `Date` header.

   - Replying now uses the new `Data.RFC5322.reply` function.  This
    results in possible different rendering of the From/To/Cc fields
    of the reply.  For example, see the UAT changes in this commit.

   - If user has not set cvIdentities, a reply will be constructed with
    the default `From` address `CHANGE.ME@YOUR.DOMAIN`.
    `Data.RFC5322.reply` requires a `NonEmpty` list of author
    mailboxes because RFC 5322 requires all messages to have a `From`
    field.  A future change could use info from the environment to
    construct a better default email, but for now we just use this
    hardcoded, in-your-face default.
```

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -149,21 +149,12 @@ import Control.Monad.Except (runExceptT, MonadError, throwError)
 import Control.Exception (IOException)
 import Data.Text.Zipper
        (insertMany, currentLine, gotoEOL, clearZipper)
-import Data.Time.Clock (getCurrentTime)
-import Data.Time.LocalTime (getZonedTime, zonedTimeToUTC)
+import Data.Time (getCurrentTime, getZonedTime)
 import System.Directory (doesPathExist)
 
 import qualified Data.RFC5322.Address.Text as AddressText
   ( renderMailboxes, addressList, mailboxList )
 import Data.MIME
-       (createMultipartMixedMessage, contentTypeApplicationOctetStream,
-        createTextPlainMessage, createAttachmentFromFile, buildMessage,
-        contentDisposition, dispositionType, headers, filename,
-        parseContentType, attachments, entities, matchContentType,
-        contentType,
-        encapsulate, MIMEMessage, WireEntity, DispositionType(..),
-        ContentType(..), Mailbox(..),
-        CharsetLookup, headerDate, headerTo, headerFrom, headerSubject)
 import qualified Storage.Notmuch as Notmuch
 import Storage.ParsedMail
        ( parseMail, getTo, getFrom, getSubject, getForwardedSubject, toQuotedMail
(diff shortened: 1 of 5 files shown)


frasertweedale:
Not to be merged until purebred-0.5 has been released, and CI is green.

src/UI/Actions.hs
```diff
@@ -1145,35 +1136,49 @@ encapsulateMail =
...
+      charsets <- use (asConfig . confCharsets)
       case mail of
         Nothing -> do
           modifying (asViews . vsFocusedView) (Brick.focusSetCurrent Threads)
           showWarning "No mail selected for replying"
         Just m -> do
           mailboxes <- use (asConfig . confComposeView . cvIdentities)
+          let
+            idents = case mailboxes of
+              [] -> pure $ Mailbox Nothing (AddrSpec "CHANGE.ME" (DomainDotAtom $ "YOUR" :| ["DOMAIN"]))
```
romanofski, line 1169: See in this case I'm still not sure what is better: pre-set text or simply nothing ... Both mean that the user is should be aware of what he's doing. Not even sure if we can send a mail with an empty `To` mailbox field now thinking of it...

src/UI/Mail/Keybindings.hs
```diff
@@ -55,7 +55,14 @@ displayMailKeybindings =
...
     , Keybinding (V.EvKey (V.KChar '?') []) (noop `focus` continue @'Help @'ScrollingHelpView)
     , Keybinding (V.EvKey (V.KChar 'r') []) (
-        replyMail
+        senderReply
+        `focus` (
+            invokeEditor ViewMail ScrollingMailView
+            :: Action 'ComposeView 'ComposeListOfAttachments (T.Next AppState)
+            )
+        )
+    , Keybinding (V.EvKey (V.KChar 'R') []) (
```
romanofski, line 64: Wasn't it `g` in mutt? Either way not a big deal ...

romanofski (approved):

src/UI/Actions.hs
frasertweedale: This is for the `From` header.  RFC 5322 requires it, hence I designed `reply` to enforce that also.  And honestly, I don't really see any case where the user wouldn't set their address(es) in the config.

src/UI/Mail/Keybindings.hs
frasertweedale: dunno... I have not touched my mutt keybindings in 15 years, but I see I have explicitly bound it to `R`.  If the mutt default is 'g' I am happy to change it.

frasertweedale: You're correct; the default is 'g': http://www.mutt.org/doc/manual/  .  I will change it.

frasertweedale:
I suppose I should get around to doing a *purebred-email* release, updating this branch and merging.

romanofski:
Yes please so we could actually do a release in the hope perhaps more people join and contribute :)

src/UI/Actions.hs
frasertweedale: I suppose if address is not set in config, when composing or replying we can prompt for it.  Alternatively we could make purebred nag the user if the config isn't set, e.g. show a warning when starting up.  In any case, I'd prefer to deal with that separately from this change.

PR #426, opened 2021-03-15
Update to Dyre 0.9

frasertweedale:
Dyre 0.9 has support for cabal store built in.  Update to 0.9 and remove our
own cabal store-related behaviour.

```
f6790da (Fraser Tweedale, 4 weeks ago)
   crash early if program was not compiled with -threaded

   It is better to crash with a clear error message than to hang when we enter
   the Brick main loop.  Dyre should take care of adding the
   '-threaded' option when compiling the custom executable, but add this check
   just in case.

32e5eda (Fraser Tweedale, 4 weeks ago)
   dyre: defaultParams -> newParams

   Avoid deprecated 'defaultParams'; use the new and improved
   'newParams'.

2346a1d (Fraser Tweedale, 4 weeks ago)
   dyre: use v0.9 to avoid work related to cabal store and -threaded

   Dyre v0.9 brings internal support for cabal store (it deduces
   -package-id from the extra include dirs).  Migrate to cabal 0.9 and remove
   our own cabal store-related behaviour.

   Dyre v0.9 also adds -threaded when the main executable is running with the
   threaded RTS.  We no longer need to explicitly provide
   '-threaded', so remove it.
```

### DIFF
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -157,10 +157,11 @@ module Purebred (
 
 import UI.App (theApp, initialState)
 
+import Control.Concurrent (rtsSupportsBoundThreads)
 import Purebred.System.Logging (setupLogsink)
 import qualified Config.Dyre as Dyre
 import qualified Control.DeepSeq
-import Control.Monad ((>=>), void)
+import Control.Monad ((>=>), void, unless)
 import Options.Applicative hiding (str)
 import qualified Options.Applicative.Builder as Builder
 import qualified Data.Text.Lazy as T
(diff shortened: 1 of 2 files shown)


frasertweedale:
CI is very grumpy about this change... have to work out why.

src/Purebred.hs
```diff
@@ -332,54 +331,25 @@ genBoundary = filter isBoundaryChar . randomRs (minimum boundaryChars, maximum b
...
-      | otherwise = []
-  in
-    packageDb <> packageId
-
 -- | Main program entry point.  Apply to a list of plugins (use
 -- 'usePlugin' to prepare each plugin for use).
 --
 purebred :: [PluginDict] -> IO ()
 purebred plugins = do
+  unless rtsSupportsBoundThreads (error "purebred was not compiled with -threaded")
```
romanofski, line 339: Ah nice!

romanofski (approved):

Issue #111, opened 2017-11-29
Creating searches based on values in the message header


romanofski:
A use case I sometimes have is finding all mails from a certain author or mailing list. Rather than crafting the query myself, I think it could be useful to have a keybinding which takes the value from the header in which it makes sense (e.g. from, to, subject) creates a query and performs the search. For multiple items it should be possible to either make the selection more finicky or have an editor to edit the selection.

PR #431, opened 2021-05-19
Search related threads by author

romanofski:
This implements a new Action which implements a related search. It picks
the first author of a thread and uses it as a full text search term.

Fixes https://github.com/purebred-mua/purebred/issues/111

### DIFF
--- a/src/UI/Actions.hs
+++ b/src/UI/Actions.hs
@@ -86,6 +86,7 @@ module UI.Actions (
   , openAttachment
   , setTags
   , saveAttachmentToPath
+  , searchRelated
 
   -- ** Actions for scrolling
   , scrollUp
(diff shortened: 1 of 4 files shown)


src/UI/Actions.hs
```diff
@@ -1321,6 +1322,19 @@ fileBrowserToggleFile =
     fb' <- lift $ FB.maybeSelectCurrentEntry fb
     assign (asFileBrowser . fbEntries) fb'
 
+-- | Search related mails sent from the same authors.
+--
+searchRelated :: Action 'Threads 'ListOfThreads ()
+searchRelated = Action ["search related mail"] $ do
+  authors <- gets (concatOf (asThreadsView . miThreads . listList . to L.listSelectedElement . _Just . to snd . to snd . thAuthors))
+  case firstOf traverse authors of
```
frasertweedale, line 1330: I think this can be written as:
```haskell
authors <- preuse (asThreadsView . miThreads . listList . to L.listSelectedElement . _Just . _2 . _2 . thAuthors . traverse)
```

frasertweedale (commented):
One nit.

UAT?

romanofski:
Yeah will add a UAT.. should be a simple one.

src/UI/Actions.hs
romanofski: Indeed!! Thanks @frasertweedale-hlint ;)

romanofski:
Alright UAT added and review comment applied.

src/UI/Actions.hs
```diff
@@ -1321,6 +1322,19 @@ fileBrowserToggleFile =
     fb' <- lift $ FB.maybeSelectCurrentEntry fb
     assign (asFileBrowser . fbEntries) fb'
 
+-- | Search related mails sent from the same authors.
+--
+searchRelated :: Action 'Threads 'ListOfThreads ()
+searchRelated = Action ["search related mail"] $ do
+  authors <- preuse (asThreadsView . miThreads . listList . to L.listSelectedElement . _Just . _2 . _2 . thAuthors . traverse)
+  case firstOf traverse authors of
+    Nothing -> runExceptT (throwError (ProcessError "No authors availabe to perform search"))
```
frasertweedale, line 1331: Not sure if `ProcessError` is suitable.  That constructor is used for subprocess errors.

romanofski: You're right.. I'm getting sloppy. Will change.

romanofski: I named it `InvalidQueryError`.. I'm not sure whether that's a bit too specific to the query. The error is perhaps more UI, than query related. If you have a better word in your mind - all ears :)

frasertweedale: `InvalidQueryError` is fine with me.  Merging.

Issue #299, opened 2019-06-17
Documentation request: enable profiling


romanofski:
**Describe the bug**
Hacking on #298 I came into the need to profile the code I wrote, but for the life of me couldn't figure out how to couple cabal and dyre to compile and run Purebred. This is not a bug report, but rather a documentation request.

**To Reproduce**
1. I run `cabal new-install exe:purebred --overwrite-policy=always --enable-library-profiling`
2. Remove the cached binary `rm -f ~/.cache/purebred/purebred-linux-x86-64`
3. Invoke purebred with `GHCOPTS="-prof" ~/.cabal/bin/purebred`
4. See error:
```
[nix-shell:~/works/purebred]$ GHCOPTS="-prof" ~/.cabal/bin/purebred 
Configuration '/home/rjoost/.config/purebred/purebred.hs' changed. Recompiling.
Error occurred while loading configuration file.
purebred: 
/home/rjoost/.config/purebred/purebred.hs:24:1: error:
    Could not find module ‘Purebred’
    Perhaps you haven't installed the profiling libraries for package ‘purebred-0.1.0.0’?
    Use -v to see a list of the files searched for.
   |
24 | import Purebred
   | ^^^^^^^^^^^^^^^

CallStack (from HasCallStack):
  error, called at src/Purebred.hs:295:32 in purebred-0.1.0.0-7f680c73d3ad41df3761c7c7966f75df65ef0457e3ac322eac46a69d3cbca90d:Purebred
```

**Expected behavior**
More documentation on how you would get a profiling purebred binary running.

**Screenshots**
None

**Additional context**
I'm currently running on NixOS. I'm happy to workaround the fact that some of the methods which work on classic Linux won't work here.

romanofski:
I've been looking into this because it has been driving me crazy that I can't simply profile stuff. I had figured it out before but just can't reproduce it any more.
My current error message is:

```
/home/rjoost/.config/purebred/purebred.hs:24:1: error:
    Could not find module ‘Purebred’
    Perhaps you haven't installed the profiling libraries for package ‘purebred-0.1.0.0’?
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
24 | import Purebred
   | ^^^^^^^^^^^^^^^

/home/rjoost/.config/purebred/purebred.hs:25:1: error:
    Could not find module ‘Purebred.Plugin.ICU’
[...]

PR #433, opened 2021-06-13
Build-flag Lazy vector to avoid bug

romanofski:
This puts a build flag around the lazy vector feature. Motivation behind
is to make the first release, get some publicity and hopefully get more
people interested to hack on Purebred. Possibly help with fixing the
bug.

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -43,6 +43,10 @@ source-repository head
   type: git
   location: https://github.com/purebred-mua/purebred.git
 
+flag with_lazyvector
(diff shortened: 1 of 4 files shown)


romanofski:
It's a draft so far.. let me know what you think about the idea.

frasertweedale:
No objection.  If you've tested both with and without flag, merge it.

romanofski:
Tested this out today with working on #299 . This is working as expected and can be enabled with `-f with_lazyvector` when building with `cabal`.

PR #438, opened 2021-09-14
nix: fix deprecation warning

lucc:
I want to  try using purebred with nix flakes.  This came up because my nixpkgs is from the unstable branch and the `stdenv.lib` does not exist there any longer.

I ran `nix-build` on this branch to check that the move to lib was already possible on the nixpkgs commit you are pinning in .nix/nixpkgs.nix and it works.

Quick question: are you interested to switch the nix support to flakes?  After all you are already implementing one of its features with nixpkgs pinning by hand.  I have not investigated any further implications but I could open a PR.


romanofski:
Great work @lucc ! Was totally overdue :) To your question: I'd totally welcome the move to nix flakes.

PR #446, opened 2021-11-01
refactor StatefulEditor

frasertweedale:
Refactor the `StatefulEditor` type and related functions.  Save the
whole `Editor` when creating a snapshot, which simplifies the
implementation.  Hide the constructor and the `_editState` field
accessor to prevent access to the saved editor state.  Add the
`statefulEditor` smart constructor which sets both the initial value
and the initial snapshot (to the same value).


romanofski:
Nice!

PR #449, opened 2021-11-03
simplify InputValidated event

frasertweedale:
The InputValidated event currently contains a Lens to the AppState
constituent that validation error messages should be written to.
But at the moment the only target is `asUserMessage`.  Furthermore,
some places use the optics whereas other places on the "sink" side
of the event channel still explicitly mention `asUserMessage` (e.g.
when clearing the message).

Simplify the type by removing the Lens argument.  Explicitly mention
`asUserMessage` in all places, eliminating the contradictions and
potential for confusion.

Perhaps we will need this behaviour again in the future - that is,
to specify the part of the AppState that a validation error shall be
written to.  If that happens, we will implement this behaviour
afresh, and hopefully avoid the aforementioned mistakes.

### DIFF
--- a/src/Purebred/Types.hs
+++ b/src/Purebred/Types.hs
@@ -1226,9 +1226,7 @@ data PurebredEvent
   | NotifyNewMailArrived Int
   | -- | Event used for real time validation. Provides the setter in
     -- the 'AppState' to set the given 'UserMessage'.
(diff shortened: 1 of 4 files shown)


romanofski (commented):
Nice little simplification really. Cheers!

PR #452, opened 2021-11-04
modules: extract several modules from Purebred.Types

frasertweedale:
From the overfull `Purebred.Types` module, extract three smaller
modules of related data types and functions:

- `Purebred.Types.Event`
- `Purebred.Types.Mailcap`
- `Purebred.Types.UI`

Also move the `TagOp` data type to `Purebre.Storage.Tags`.


romanofski:
\o/

PR #454, opened 2021-11-08
avoid unnecessary recompilation due to CPP

frasertweedale:
CPP is currently used in several large modules to choose between
lazy or strict list representations.  This leads to unecessary
recompilation when `cabal_macros.h` changes (e.g. due to dependency
changes).  See blog post for further explanation:

http://blog.haskell-exists.com/yuras/posts/stop-abusing-cpp-in-haskell.html

Abstract the list type behind a new module `Purebred.Types.Items`,
which uses CPP.  Refactor other modules that use CPP to choose a
list representation to use `Items` and avoid CPP in those modules.

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -94,6 +94,7 @@ library
                      , Purebred.Version
   other-modules:       Paths_purebred
                      , Purebred.Types.IFC
+                     , Purebred.Types.Items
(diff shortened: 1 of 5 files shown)


romanofski (commented):
I see.. nice refactoring!

PR #456, opened 2021-12-21
nix: stick with building purebred for x86 for now

romanofski:
The flake-utils allow us to build purebred for non-x86-64 systems.
However at the beginning we don't support these platforms. Part of the
downside is that we're also snapping up any potential broken package in
the nixos-unstable channel. Granted, we could pin to a specific
revision, but the channel is mostly fine for x86_64.

Drive-by add on to also run `nix flake check` in CI.

### DIFF
--- a/.github/workflows/nixos-ci-flakes.yml
+++ b/.github/workflows/nixos-ci-flakes.yml
@@ -1,0 +1,24 @@
+name: NixOS Build (Flakes - build only)
+on:
+  - push
+  - pull_request
+jobs:
+  linux:
+    name: Nix
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


romanofski:
```
HLint version 3.3.5
Downloading HLint version 3.3.5
Status Code: 404 Redirects: 0 Total time: 0.172183 Total Dsize: 9

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```
The archive isn't there... will have to figure out why ...

frasertweedale:
Never mind - it's not related to this change.  Merging.

PR #460, opened 2022-06-11
ci: Fix nix build error

romanofski:
This adds the newer version of `typed-process` into the package mix,
since just updating to a newer package set will clash with our version
pins in `hs-notmuch` and `purebred-email` respectively.


frasertweedale:
Took a few goes but CI is green.  Merging.

PR #462, opened 2022-07-03
Refactor and simplify nix sources

romanofski:
We currently define sources to build purebred in nix in different
files. Both keep around redundant information found in both. Since nix
flakes is still experimental, we didn't want to make `nix develop` the
default when building the nix package. So we kept the older `nix-build`
around which also defined overlays etc.

Instead simplify everything by defining a single source of truth in
`overlays` which is then used by two methods of building in nix: the old
`nix-build` way and the newer `nix build`.

### DIFF
--- a/.github/workflows/nixos-ci.yml
+++ b/.github/workflows/nixos-ci.yml
@@ -9,12 +9,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Local cache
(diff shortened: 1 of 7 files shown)


frasertweedale:
Nix is your department so I'll just wave this through :)

romanofski:
Oh @frasertweedale you were too quick with this one :D

PR #465, opened 2022-07-21
CI and UAT improvements

frasertweedale:
- add GHC 9.2 and 9.4 to CI matrix
- apply an hlint suggestion
- (UAT) update for some error message changes in GHC >= 9.2
- (UAT) fix a GNU-ism in test setup (`cp` command)
- (UAT) set `TERM=screen` instead of `TERM=ansi` which lacks some capabilities on FreeBSD

Note that UAT still isn't green on GHC 9.2 and more needs to be done.  8.8, 8.10 and 9.0 should be green.


test/TestUserAcceptance.hs
```diff
@@ -1817,10 +1809,15 @@ setUp (GlobalEnv globalConfigDir) sessionName = do
   runProcess_ $ proc "sh" ["-c", "cp -a " <> globalConfigDir <> "/* " <> confdir]
 
   flip runReaderT sessionName $ do
-    -- a) Make the regex less color code dependent by setting the TERM to 'ansi'.
+    -- a) Make the regex less color code dependent by setting the TERM to 'screen'.
```
romanofski, line 1812: Yeah nice!

romanofski:
btw ... I presume you want to fix the Haskell CI errors? I'm still on getting those nix errors fixed... not really sure why the normal Nix build is giving me a different outcome.

frasertweedale:
@romanofski I want to merge this as-is, if you're happy with it.  The 9.2 failures are a bit nasty and may take some time to solve.

PR #466, opened 2022-07-21
Refactor and simplify nix sources

romanofski:
We currently define sources to build purebred in nix in different
files. Both keep around redundant information found in both. Since nix
flakes is still experimental, we didn't want to make `nix develop` the
default when building the nix package. So we kept the older `nix-build`
around which also defined overlays etc.

Instead simplify everything by defining a single source of truth in
`overlays` which is then used by two methods of building in nix: the old
`nix-build` way and the newer `nix build`.


frasertweedale:
@romanofski merge this when you feel it's ready.  I lack the knowledge of nix to provide any feedback at all.

romanofski:
I disabled the normal nix build for now causing cabal install to not be able to find `word-wrap` even tho it's built. Not exactly sure what causes the difference to the new method of building nix packages (nix flake), but I'm also a bit over it so I merge this now and may have a look into this later.

PR #470, opened 2022-07-23
refactor: remove Toggleable from db functions

frasertweedale:
The current database query functions return lists of Toggleable
thread/message.  But this concept doesn't belong in the database
API.  Push the construction of the Toggleable value down to the UI
code, applied to the result of the database queries.


romanofski:
There are two tests which seem to fail consistently - you want to investigate?

frasertweedale:
@romanofski the only jobs that are failing are the GHC 9.4 tests (allowed to fail; caused by vty not supporting GHC 9.4 yet), and the NixOS jobs.  Apart from that it's just the usual test flakiness.

Issue #471, opened 2022-07-28
RFC: decompose toggled/selected actions


frasertweedale:
Within the UI code, we have `toggledOrSelectedItemHelper` for several actions,
including:

- `updateReadState`
- `completeMailTags`
- `setTags`

The helper has the following behaviour:

- If there are any items "toggled" on, then the action is applied to the **toggled** items (bool set to true in the action item)
- Otherwise, the action is applied to the **selected** item (whatever item the Brick list cursor is on)

I propose that we "decompose" this action into separate actions/keybindings for "apply to toggled items"
and "apply to current item".  This decomposition would then enable a clean-up of the `HasList` and `HasToggleableList`
classes and instances, providing for more ergonomic access and safer handling of items in toggleable lists.
For example, we can avoid exposing the "toggled" state in traversals that targets toggled items.

It may also make some things in the UI less ambiguous as to what items a given operation shall be applied to.

Seeking your feedback @romanofski on this idea.

romanofski:
Sounds good. The current state is a "make-it-work" implementation anyways. I guess it would also be useful due to the changes in Brick in the background?

PR #472, opened 2022-08-02
Refactor HasToggleableList, and other cleanups

frasertweedale:
Related to https://github.com/purebred-mua/purebred/issues/471, and accomplishes some of the
refactors and cleanup I had in mind when filing that issue - but without any behavioural changes
(hopefully!)

```
commit 20b614ad01a92127b0ab45ab365cc761cf72c943
Date:   Thu Jul 28 16:42:34 2022 +1000

    ui: simplify 'toggle'
    
    Simplify the Toggleable class and in particular the 'toggle' effect.
    The previous approach was "two phase" - first find the index of the
    current item, then toggle it.  We now avoid this by way of the
    'listSelectedElementL' Traversal, which targets the currently
    selected item.
    
    `listSelectedElementL` has been merged into Brick itself, but is not
    yet released.  We can migrate to the Brick-supplied version after it
    has been released.

commit b5b4f328c81adccc9c3b6bd37d8b1a6d54dd3f28
Date:   Fri Jul 29 11:34:57 2022 +1000

    ui: toggledItemsL: target "inner" values
    
    The toggledItemsL targets the whole `Toggleable a`, which includes
    the "is toggled" `Bool`.  This makes it an unlawful `Traversal` -
    you can changed the state that says which elements are targeted.  It
    is also unneeded.  Functions are forced to handle a (Bool, a), but
    always discard the first element.
    
    Fix these problem by changing 'toggledItemsL' to target the inner
    `a` values of the `Toggleable a` list items.  This makes it a lawful
    `Traversal` and simplifies the traversal functions.

commit 2e5257862c5f5dce21b9c0538d37531322e5c494
Date:   Tue Aug 2 22:01:31 2022 +1000

    ui: use type applications instead of Proxy
    
    Make things just a little tidier in our Purebred.UI.Actions by using
    -XTypeApplications.  This avoids the need to construct and
    explicitly pass around Proxy values.

commit 431d0e44fb7fe20fb35edcfedca98dead306574b (HEAD -> refactor/toggleable, origin/refactor/toggleable)
Date:   Tue Aug 2 22:08:24 2022 +1000

    ui: extract untoggleAll and toggledItemsL from class
    
    The `untoggleAll` and `toggledItemsL` functions are defined in terms
    of other members of `HasToggleableList` and its superclass
    `HasList`.  Extract them from the class as top-level values.
```

### DIFF
--- a/src/Purebred/UI/Actions.hs
+++ b/src/Purebred/UI/Actions.hs
@@ -15,7 +15,6 @@
 -- along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 {-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -123,7 +122,6 @@ import qualified Brick.Widgets.FileBrowser as FB
 import Brick.Widgets.Dialog (dialog, dialogSelection, Dialog)
 import Network.Mime (defaultMimeLookup)
 import Data.Kind (Type)
-import Data.Proxy
 import Data.Either (fromRight, isRight)
 import qualified Data.Text as T
 import qualified Data.Text.Lazy as LT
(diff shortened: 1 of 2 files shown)


src/Purebred/UI/Actions.hs
```diff
@@ -216,114 +214,138 @@ first. Keep the 'Action' specific to the view and widget
...
-  isToggledE _ = fst
-  toggle proxy i = modifying (list proxy . L.listElementsL . ix i) (toggleE proxy)
+  type Inner n = Snd (E n)
+  toggleState = _1
+  inner = _2
+  toggle = modifying (list @n . listSelectedElementL . toggleState @n) not
+
+-- | Traversal of selected element (if any)
+--
+-- We should try to upstream this to Brick
```
frasertweedale, line 333: Actually it was already merged: https://github.com/jtdaugherty/brick/pull/382

I mentioned in the commit message but forgot to remove/update this comment.  I'll fix it if I have to rebase or revise this PR.

frasertweedale:
*looks at ci*

welp.  I thought this should be a mechanical refactor with no behavioural change.  Obviously I was wrong.  Converting to draft until I discover why it broke (or give up).

frasertweedale:
OK, found and fixed the errors.  Ready for review.

romanofski:
Good stuff!

PR #474, opened 2022-08-04
more refactors (move some definitions)

frasertweedale:
```
commit c936d6ec7767045fe48108ecd604b41bcc5faee7 (HEAD -> refactor/moves, origin/refactor/moves)
Date:   Fri Aug 5 00:33:02 2022 +1000

    move getDatabasePath to Purebred.Config
    
    Let's keep Purebred.Storage.* focused on the database operations
    rather than finding out where the database is, which is a config
    consideration.

commit 3c407e436dccbd44b9293b52fbb0d115254475b2
Date:   Fri Aug 5 00:32:00 2022 +1000

    extract tag management functions to Purebred.Storage.Tags
```


romanofski (approved):
\o/

PR #475, opened 2022-08-05
nix: Make purebred with ICU the default package

romanofski:
Currently the purebred default package in the nix flake is set to
purebred without ICU. But arguably the ICU plugin is something most
users would want to have hence this patch to make it a default package.


frasertweedale:
ACK I guess - merge when you're ready :)

Issue #483, opened 2022-09-26
decide versioning scheme


frasertweedale:
**Is your feature request related to a problem? Please describe.**

We should claim the /purebred/ package name on hackage.
To do that we need a release (or a candidate release, to at least claim the package name).
Before we do that we should decide upon a versioning scheme.

**Describe the solution you'd like**

Releases will be sporadic and will usually break compatibility in some way (of the library, the configuration data type, or the behaviour of the program).   I suggest the following scheme:

> `<year>.<releasenum>[.<patchnum>]`, e.g. `2022.1` for the first release in 2022, `2022.2` for the second release in 2022, and `2022.n.p` for bugfix releases that don't break API or config in any way. 

This scheme is PVP and semver compliant.  If that even matters for a user application (as opposed to library).  But I reckon it does to some extent, given that we want to encourage plugins.

**Describe alternatives you've considered**

Instead of year we could just increment the first version component for each "major" release.  E.g. `1.0` then `2.0` and so on, with patch level releases being `n.0.p`.  This scheme has an unused second version component (for PVP compliance).

Or we could use `0.n.p` or `1.n.p` and stay at "version 0" or "version 1" until we decide to become "the next major version".  This scheme avoids unused version components.  But it has some mysterious meaning ascribed to the first version component.  What circumstances warrant a bump from `v.n[.p]` to `(v+1).0`?  It might be better to avoid the question :)

frasertweedale:
@romanofski let me know your thoughts.

romanofski:
Yeah I was wondering do we want to expose that it's still work in progress or just release 1.0 and say good enough and use the normal `major/minor/patch` scheme? Or in other words, if we use `<year>.<releasenum>[.<patchnum>]` would we want to stick with it or just until we release `1.0`?

frasertweedale:
@romanofski I want to decide the permanent scheme.  We can stick with 1.x etc (fair enough, we already defined milestones as such).  But I was thinking about the meaning (or lack of) implied by different schemes and what actually makes the most sense.  So, just let me know what you think works best or provide your thoughts on the advantages/disadvantages of the different options.

romanofski:
Ah understood. I agree with the `<year>.<releasenum>[.<patchnum>]` conveying a bit of a slow release cycle and I think that's a good idea. Renaming our current release milestones is simple, so I don't think we should be holding off on any different versioning scheme really.

[...]

PR #484, opened 2022-09-27
become v2022.1

frasertweedale:
- ci updates
- README and `purebred.cabal` doc updates
- become version `2022.1`, ahead of first release to hackage

Related issue: https://github.com/purebred-mua/purebred/issues/483


frasertweedale:
Note: GHC 9.4 build fails because of `base < 4.18.0.0` upper bound in `brick`.  See https://github.com/jtdaugherty/brick/issues/396

Issue #448, opened 2021-11-03
mail view: search within body does not scroll to search term


frasertweedale:
**Describe the bug**

When viewing a mail, searching for a term within the body of a mail, where that term appears multiple times,
scrolls down 1 line each time when calling the "next" occurrence.

**Expected behavior**
Viewport should scroll to put the "next" occurrence of the search term in view.

romanofski:
I'll keep it with in the 1.0 release. Since shipping a b0rked search would be confusing and diminish the usefulness.

frasertweedale:
Assigning to me; I have a fix.

PR #485, opened 2022-09-28
fix mail view search hit scrolling

frasertweedale:
"Next search hit" scrolling in mail view is busted.  The main issue
is that we were scrolling to line *n* where *n* is the search hit
*number*, rather of the *line number* of the search hit.  Reference
field `_2` instead of `_1` to resolve this issue.

A secondary issue is scrolling did not account for inter-paragraph
padding.  As a result, the scroll point diverges from the search hit
at a rate of one line per paragraph.  With sufficiently many
paragraphs the search hit is not in the viewport.  To resolve this
issue, eliminate the whole `[Paragraph]` type.  Use a flat `[Line]`
instead.

Also eliminate the "line number" field from the `Line` data type.
Instead, use an indexed traversal to reveal the line number when
searching for occurrences of the search string.  This change further
simplifies `MailBody` construction and avoids storing unnecessary
data.

Fixes: https://github.com/purebred-mua/purebred/issues/448


romanofski:
Aaah sweet. Very nice!! Thank you!

PR #487, opened 2022-09-29
mail body search: refactors, add "previous match" action

frasertweedale:
```
9206083 (Fraser Tweedale, 9 minutes ago)
   add scrollPreviousMatch action

   Alongside the existing "scroll to next match" action, add the
   `scrollPreviousMatch` action.  Bind `'N'` to this action in the default
   keybindings.

   Also rename `scrollNextWord` to `scrollNextMatch` which better expresses
   the real behaviour (a search term can match part of a term or span word
   boundaries).

d68f2f2 (Fraser Tweedale, 49 minutes ago)
   further simplify mail body search

   The remains a considerable dupliation of data within our implementation of
   mail body search.  In particular, `ScrollStep` duplicates `Match` values
   from `Line`.

   Refactor the implementation in the following ways:

   - Promote the `[Match]` from the `Line` data type to the `MailBody`
    type (as a single, combined list)

   - Remove the `Line` data type, replacing it with bare `Text`

   - Replace the `FocusRing ScrollStep` with an `Int` that records the
    current search hit index.

   - Extract the "scroll to search index" behaviour as a helper action.
    It modifies the search index, wrapping if necessary.  Update
    relevant actions to use the new method.

   - Clean up the rendering function
```

### DIFF
--- a/src/Purebred/UI/Actions.hs
+++ b/src/Purebred/UI/Actions.hs
@@ -93,7 +93,8 @@ module Purebred.UI.Actions (
   , scrollDown
   , scrollPageUp
   , scrollPageDown
-  , scrollNextWord
+  , scrollNextMatch
+  , scrollPreviousMatch
   , removeHighlights
 
   -- ** Actions for composing mails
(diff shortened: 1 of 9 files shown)


src/Purebred/UI/Actions.hs
```diff
@@ -1015,23 +1015,49 @@ scrollPageDown = Action
...
+
+scrollPreviousMatch :: forall ctx v. (Scrollable ctx) => Action v ctx ()
+scrollPreviousMatch =
+  Action
+    { _aDescription = ["scroll to previous match in mail body"]
+    , _aAction = scrollMatch (makeViewportScroller @ctx) (subtract 1)
     }
 
+-- | Scroll to a particular match, the index of which can be calculated
+-- from the current index.  So:
```
romanofski, line 1033: This basically scrolls relatively to a match if I read this correctly instead of what I used to be relative to the top (e.g. line 0)?

romanofski (commented):
Good stuff! Thank you.

src/Purebred/UI/Actions.hs
frasertweedale: Under the hood, it still does "scroll to beginning, then scroll forward <n> lines".  But now we calculate <n> a bit differently.

PR #490, opened 2022-10-07
abstract EntityCommand over the output type

frasertweedale:
`EntityCommand` final output type was `T.Text`.  The `ccAfterExit`
must convert the raw command output from `Tainted L.ByteString` to
`T.Text`.  This was not ideal.  In general, features that execute
commands over body content might need the raw bytes, or might wish
to convert the raw bytes into something other than `T.Text`.

Therefore this commit makes the following changes:

- Add an output type parameter to `EntityCommand`.

- Equip `EntityCommand` with a `Functor` instance, which modifies
  the output value/type.

- Generalise the `ccAfterExit` optic to allow changing the output
  type.

- `handleExitCodeThrow` and `handleExitCodeTempfileContents` now
  return `Tainted L.ByteString`.

- Extract a helper function to avoid duplicate error handling code
  in `handleExitCodeThrow` and `handleExitCodeTempfileContents`.

- Update usage sites, including performing the conversion to
  `T.Text` that was previously done by the `ccAfterExit` value.

- For convenience, add the `EntityCommandAfterExit` type synonym.

- Remove unneeded constraints from the `ccResource` and
  `ccRunProcess` lenses.

This refactor will facilitate an upcoming generalisation of the
"mailcap" configuration.  It also enables more kinds of body
processing in the future.

### DIFF
--- a/src/Purebred/Storage/Mail.hs
+++ b/src/Purebred/Storage/Mail.hs
@@ -65,7 +65,7 @@ import Purebred.Types.Mailcap
   , mailcapHandlerToEntityCommand
   )
 import Purebred.Storage.Client (Server, mailFilepath)
-import Purebred.System.Process (runEntityCommand)
+import Purebred.System.Process (outputToText, runEntityCommand)
 
 {- $synopsis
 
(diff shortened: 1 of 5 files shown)


romanofski (approved):
Nice change!

PR #493, opened 2022-10-20
ui: improve mail/thread index field alignment

frasertweedale:
@romanofski give this a go and see if you think it's an improvement.


romanofski:
Is the main big difference that threads with lots of replies (e.g. more than 10) align properly now, where as before they the tags and subject would be pushed to the right (by how many digits the amount of replies are)?

frasertweedale:
Pretty much.  Thread message count is now right-aligned, and I fixed a discrepancy in padding, so everything lines up.

frasertweedale:
Oh, and I also deleted the space char at the start of each line.  Not sure if you really wanted it there but I'd prefer 1 char additional content.

romanofski:
Nah, I love it. Great stuff @frasertweedale. I don't get that many long threads, but I presume for you it's much much more pronounced.

Issue #492, opened 2022-10-20
tag replacement in thread and mail list


frasertweedale:
**Is your feature request related to a problem? Please describe.**

Tags (e.g. `inbox`, `replied`, `attachment`, `flagged`, etc) take up significant horizontal space in the Thread and Mail list.

**Describe the solution you'd like**

I would like a (customisable) facility to replace particular tags with a shorter string e.g. a single char or emoji.  For example, `replied` replaced with ↩️, `attachment` with 📎, and so on.  A replacement must allow more than 1 `Char`, to allow for combining/modifier codepoints.  We can ship with a default set or maybe two (one ASCII, one emoji) of replacements for [special tags](https://notmuchmail.org/special-tags/).

**Describe alternatives you've considered**
Do nothing and put up with loss of horizontal space for actual subject.

romanofski:
Yeah I've seen those translations happen in other editors. I wonder if a naive implementation is just a map we can look up the icons for each tag. Could perhaps be stored in the Config?

frasertweedale:
@romanofski I already have a branch.  Will push it soon.

PR #494, opened 2022-10-22
ui: enable tag replacement in thread and mail list

frasertweedale:
Add a "tag replacement map" config.  Update tags widget rendering to
gather the replacements and render them together, ahead of remaining
tags.  This enables users to define symbolic replacements for
particular tags.

The default config replaces `flagged`, `attachment` and `replied`
with emoji.  We also export and ASCII variant of this map.  Users
can extend or replace the map via the `TweakConfig` plugin.

Fixes: https://github.com/purebred-mua/purebred/issues/492

### DIFF
--- a/src/Purebred/Config.hs
+++ b/src/Purebred/Config.hs
@@ -28,6 +28,7 @@ import Control.Lens (set)
 import Control.Monad.Except (runExceptT)
 import qualified Data.ByteString.Builder as B
 import qualified Data.ByteString.Lazy as L
+import qualified Data.Map as M
 import qualified Data.Text as T
 import System.Directory (getHomeDirectory)
 
@@ -208,6 +209,7 @@ defaultConfig = do
       , _ivFromKeybindings = gatherFromKeybindings
       , _ivToKeybindings = gatherToKeybindings
       , _ivSubjectKeybindings = gatherSubjectKeybindings
+      , _ivTagReplacementMap = tagReplacementMapEmoji
       }
     , _confComposeView = ComposeViewSettings
       { _cvFromKeybindings = composeFromKeybindings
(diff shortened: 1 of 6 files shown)


src/Purebred/Config.hs
```diff
@@ -234,3 +236,36 @@ defaultConfig = do
...
+-- | Replace some special tags with emoji.
+--
+-- * @flagged@ → 📌
+-- * @attachment@ → 📎
+-- * @replied@ → ↩️
+--
+tagReplacementMapEmoji :: M.Map T.Text T.Text
+tagReplacementMapEmoji = M.fromList
+  [ ("flagged", "📌")
+  , ("attachment", "📎")
```
romanofski, line 262: can we add `inbox` and `archive` too? :D

frasertweedale: `archive` doesn't seem to have any special meaning to notmuch: https://notmuchmail.org/special-tags/.
So I'm inclined to omit it.  You can add it via custom config.

`inbox` sure, what emoji and ascii char do you want?

romanofski: What do you think about: :inbox_tray: and `I` simply as an ASCII representation. Suppose we would want to keep the ASCII representation somewhat memorable.

frasertweedale: OK, PR updated, give it a spin.

An alternative @inbox@ emoji would be https://emojipedia.org/incoming-envelope/ but I'll let you decide.

frasertweedale:
@romanofski it seems that the emoji width causes some problems in the test environment.  I'll switch to using the ASCII replacement map in our test config.  But I won't get to that right away so this PR is on hold for now.

Feel free to continue testing / playing with the new feature.

romanofski:
> @romanofski it seems that the emoji width causes some problems in the test environment. I'll switch to using the ASCII replacement map in our test config. But I won't get to that right away so this PR is on hold for now.
> 
> Feel free to continue testing / playing with the new feature.

Tried it out. I like it!

frasertweedale:
@romanofski I fixed the tests.  If you're happy with this change, :ship:

PR #495, opened 2022-11-18
Since vty-5.34 Vty.Event has a NFData instance

mlang:


frasertweedale:
Thanks @mlang!  Re-running the flaky tests.  I'll merge after everything passes.

Issue #17, opened 2017-07-21
Addressbook


romanofski:
We need a way to store and retrieve addresses. This might also be something which could go into a plugin, while the most basic form could be simply parsed from a file similar to what mutt does. 

Thinking of it, it would be nice if there are multiple forms addresses can be looked up. Just because you might keep some addresses local, but have access to an ldap server.

frasertweedale:
The challenge here that composing mail will be done (of course) by invoking
`$EDITOR` unto a tempfile.  So we first need to decide if it is in scope or not.

I guess we can make it a first step separate from the editing, or edit *just* the body
and have purebred handle all addressing.

In which case we want something like:

```haskell
type AddressBook = AddressSearchTerm -> IO [RFC822Address]
```
Then it is easy to have multiple AddressBooks; just traverse them, merge the results and present to user.

romanofski:
I was thinking of mutt, in which the workflow goes:

* enter subject
* invoke editor on tempfile in which you write the body, save, quit
* final mail view in which you change `To:`, `Bcc:`, `CC:`, etc and the attachments

The final point allows you to query different address books with different commands. I always perceived it as a good workflow, however I might have just learned it the way it is. You think this can be improved?

romanofski:
Notmuch has now a way to retrieve addresses based on search terms from the existing mail corpus. We should definitely use that as a first class citizen. Perhaps that alleviates to keep some kind of addressbook file. I guess to support other address retrieval (e.g. LDAP), that could be implemented in a plug-in.

romanofski:
I looked only in the CLI, since I thought being able to search the addresses in Xapian shouldn't be much of a problem. I did not check how it was implemented.
I think having a small tool to populate an address book might already be enough if it really is just the function you were referring to.

frasertweedale:
@romanofski the `notmuch address` command:

1. does a xapian search for all messages matching the query, and iterates them
2. if `--output=sender` is given, the file has to be opened and parsed to read the To, Cc and Bcc headers.  Otherwise only the From header is looked at (which is indexed, so file does not have to be parsed).
3. stuff the addresses into a hashtable and output the result.\

So, it is not efficient as part of an automatic address book population.  Separate small tool, sure.

romanofski:
What do we settle as an acceptance criteria for this item? Implement a way to retrieve an address based on an IO action, which will default to retrieve the address from a text file? As a text file, lets use the mutt address file format?

[...]

PR #499, opened 2022-12-11
Allow to expand receipients from addressbook

romanofski:
This implements an addressbook API and a way to expand addresses based on addressbook entries. The default retriever is based on the Mutt `alias` file using nick names.

The retrieval of Mutt aliases is not enabled by default, due to the fact that we need to use an absolute path to the alias file. Guessworking the possible location of the file goes against the current nature of being very explicit in the configuration.

Fixes https://github.com/purebred-mua/purebred/issues/17

### DIFF
--- a/src/Purebred/Storage/AddressBook.hs
+++ b/src/Purebred/Storage/AddressBook.hs
@@ -1,0 +1,47 @@
+-- This file is part of purebred
+-- Copyright (C) 2022 - Róman Joost
+--
+-- purebred is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+module Purebred.Storage.AddressBook
+  ( -- * Synopsis
+    -- $synopsis
+
+    -- * API
+    queryAddresses
+  )
+where
+
+import qualified Data.Text as T
+
+import Control.Lens
+import Control.Monad.Except (MonadError)
+import Control.Monad.IO.Class (MonadIO)
+
+import Data.IMF (Address)
+
+import Purebred.Types.Error
+import Purebred.Types.AddressBook
+import Purebred.System (tryIO)
+
+-- $synopsis
+--
+-- This module provides an API to retrieve Address information for Purebred
+
+-- | query addresses matching given needle
+--
+queryAddresses :: (MonadError Error m, MonadIO m) => T.Text -> [AddressBook] -> m [Address]
+queryAddresses needle = fmap concat . traverse (tryIO . flip (view addressBookSearch) needle)
--- a/src/Purebred/Storage/AddressBook/MuttAliasFile.hs
+++ b/src/Purebred/Storage/AddressBook/MuttAliasFile.hs
@@ -1,0 +1,95 @@
+-- This file is part of purebred
+-- Copyright (C) 2022 - Róman Joost
+--
+-- purebred is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU Affero General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+module Purebred.Storage.AddressBook.MuttAliasFile
+  ( -- * Synopsis
+    -- $synopsis
+
+    -- * API
+    initMuttAliasFileAddressBook
+
+    -- * Muttalias Parser
+  , MuttAlias(..)
+  , muttAliasNick
+  , muttAliasAddress
+  , parseMuttAliasFile
+  ) where
+
+import Data.Attoparsec.ByteString (Parser, parseOnly, sepBy, string, takeTill)
+import Data.Attoparsec.ByteString.Char8 (endOfLine, isSpace_w8, skipSpace, space)
+import Data.Bifunctor (bimap)
+import qualified Data.ByteString as B
+import qualified Data.Text as T
+import qualified Data.Text.Internal.Search as T
+import Control.Lens (Lens', lens, toListOf, folded, filtered, view)
+
+import Data.MIME (defaultCharsets)
+import Data.IMF (address, Address)
+
+import Purebred.Types.Error (Error(ParseError))
+import Purebred.Types.AddressBook (AddressBook(..))
+import Purebred.Types.String (decodeLenient)
+
+-- $synopsis
+--
+-- This module provides functions to retrieve addresses from an alias
+-- file formatted in an mutt-alias compatible format.
+--
+-- https://gitlab.com/muttmua/mutt/-/wikis/MuttGuide/Aliases
+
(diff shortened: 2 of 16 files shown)


src/Purebred/Storage/Addressbook.hs
```diff
@@ -0,0 +1,87 @@
...
+queryAddresses :: (MonadError Error m, MonadIO m) => T.Text -> [AddressRetriever] -> m [Address]
+queryAddresses needle retrievers =
+  traverse (go needle) retrievers >>= pure . concat
+  where
+    go n r =
+        let config = view arConfig r
+        in liftIO (view arRetrieve r config n)
+
+-- | XXXX dependency on the constructor
+runMuttAliasRetriever :: (MonadIO m) => AddressRetrieverConfig -> T.Text -> m [Address]
```
romanofski, line 64: @frasertweedale I hate this, since it makes it hard to expand. I wonder whether I'm missing something in the type 🤔 Would you have a suggestion what to improve here?

romanofski:
Btw. I already had a listing view implemented to browse the addresses, but removed it in this patch because it made the patch bigger than it should be. Thought it's probably better to put that as a separate optimization for addresses.

src/Purebred/Storage/Addressbook.hs
frasertweedale: @romanofski what do you mean by "makes it hard to expand"?

src/Purebred/Types.hs
```diff
@@ -438,6 +448,14 @@ fbSearchPathKeybindings = lens _fbSearchPathKeybindings (\cv x -> cv { _fbSearch
 fbHomePath :: Lens' FileBrowserSettings FilePath
 fbHomePath = lens _fbHomePath (\s a -> s { _fbHomePath = a })
 
+data AddressbookSettings = AddressbookSettings
```
frasertweedale, line 451: [nit] can we capitalise `Book` i.e. `AddressBookSettings`?

src/Purebred/Types/Addressbook.hs
```diff
@@ -0,0 +1,73 @@
...
+data MuttAlias = MuttAlias
+  { _maNick :: T.Text,
+    _maAddress :: Address
+  } deriving (Show, Eq)
+
+maNick :: Lens' MuttAlias T.Text
+maNick = lens _maNick (\m x -> m { _maNick = x })
+
+maAddress :: Lens' MuttAlias Address
+maAddress = lens _maAddress (\m x -> m { _maAddress = x })
```
frasertweedale, line 73: The Mutt-specific stuff should be extracted to a separate file - as should each specific address book implementation.  e.g. `Purebred.AddressBook.MuttAliasFile`

frasertweedale (commented):
Wow, nice work mate.  Looking forward to testing it.  I left some comments.

src/Purebred/Storage/Addressbook.hs
romanofski: As in being extensible. If one of the goals was to provide an extensible API I can foresee problems with the data types if I for example would want to add support for an LDAP retriever. Since the constructor might be different and while pattern matching would work, it feels as if we used something more flexible. Yet I can't think of what it was.

src/Purebred/Types/Addressbook.hs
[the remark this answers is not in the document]
romanofski: Ah yeah fair enough. It could make some very lengthy, but I suppose better than guessing around.

src/Purebred/Storage/Addressbook.hs
frasertweedale: Thanks @romanofski, I understand.  So, I think my comment at https://github.com/purebred-mua/purebred/pull/499#discussion_r1045429227 addresses this concern.  The idea is that the address retriever UI is "just" a `Text -> m [Address]`, and you give it its configuration when constructing it.  Then all address book implementations have a simple and uniform query API.

romanofski: Yeah that is definitely much better. I totally forgot about the possibility of partials (doh!). I like this much better now.

src/Purebred/Types/Addressbook.hs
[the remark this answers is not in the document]
romanofski: Alright changed. I did leave the abbreviations for the general types tho. Thought it might be better to address this with some clean up work.

src/Purebred/Types/Addressbook.hs
```diff
@@ -0,0 +1,73 @@
...
+
+import Control.Lens (Lens', lens)
+import qualified Data.Text as T
+import GHC.Generics (Generic)
+import Control.DeepSeq (NFData(rnf), force)
+
+import Data.IMF (Address)
+
+data AddressRetrieverConfig = MuttAliasConfig FilePath | LDAPConfig String
+  deriving (Generic, NFData)
```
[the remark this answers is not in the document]
romanofski: Yes.. much much better. Thanks for the hint! Much appreciated.

src/Purebred/Types.hs
romanofski: Yep, done!

src/Purebred/Types/Addressbook.hs
romanofski: No worries. Moved.

src/Purebred/Types/AddressBook.hs
```diff
@@ -0,0 +1,68 @@
...
+
+import Control.Lens (Lens', lens)
+import qualified Data.Text as T
+import GHC.Generics (Generic)
+import Control.DeepSeq (NFData(rnf))
+
+import Data.IMF (Address)
+
+data AddressRetrieverConfig = MuttAliasConfig FilePath | LDAPConfig String
+  deriving (Generic, NFData)
```
frasertweedale, line 46: This type is now unused and should be removed.

src/Purebred/Types/AddressBook.hs
```diff
@@ -0,0 +1,68 @@
...
+data MuttAlias = MuttAlias
+  { _muttAliasNick :: T.Text,
+    _muttAliasAddress :: Address
+  } deriving (Show, Eq)
+
+muttAliasNick :: Lens' MuttAlias T.Text
+muttAliasNick = lens _muttAliasNick (\m x -> m { _muttAliasNick = x })
+
+muttAliasAddress :: Lens' MuttAlias Address
+muttAliasAddress = lens _muttAliasAddress (\m x -> m { _muttAliasAddress = x })
```
frasertweedale, line 68: We should move these definitions to `Purebred.Storage.AddressBook.MuttAliasFile`.

frasertweedale (commented):
Thanks for the update.  A few nits.  There's also a API change I'd like to make but I feel it would be best to attack it while pair programming, if you're free some time soon.

src/Purebred/Storage/AddressBook.hs
```diff
@@ -0,0 +1,49 @@
...
+import Purebred.Types.AddressBook
+-- $synopsis
+--
+-- This module provides an API to retrieve Address information for Purebred
+
+-- | query addresses matching given needle
+--
+queryAddresses :: (MonadError Error m, MonadIO m) => T.Text -> [AddressRetriever] -> m [Address]
+queryAddresses needle retrievers =
+  traverse (go needle) retrievers >>= pure . concat
```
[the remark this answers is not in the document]
romanofski: Hah.. I just hacked on it. Will adjust.

src/Purebred/Types/AddressBook.hs
romanofski: moved

src/Purebred/Types/AddressBook.hs
romanofski: ah of course.. removed

src/Purebred/Storage/AddressBook/MuttAliasFile.hs
```diff
@@ -0,0 +1,73 @@
...
+runMuttAliasRetriever :: (MonadIO m) => FilePath -> T.Text -> m [Address]
+runMuttAliasRetriever filePath needle = do
+  contents <- liftIO $ B.readFile filePath
+  let result = parseMuttAliasFile contents
+  case result of
+    Left _ -> pure []
+    Right ax -> pure $ filterMuttAliases needle ax
+
+filterMuttAliases :: T.Text -> [MuttAlias] -> [Address]
+filterMuttAliases substr ma = toListOf (folded . filtered (\x -> matchesSubstring substr (view muttAliasNick x)) . muttAliasAddress) ma
```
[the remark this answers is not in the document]
romanofski: :D done

frasertweedale:
@romanofski looking good mate.  Gonna reach out to see if we can find a time to pair to discuss one more change.

romanofski:
Just noticed that this is b0rked somewhat. Works "well" if all you do is just retrieve a single address, but if you need a few more you're in trouble since it replaces the entire line.
Will definitely have to fix that.

Also needs to be registered for the action when we forward mail and probably the `CC` and `BCC` fields.

romanofski:
Alright updated. This should do now. I parse out the addresses, grab the tail after the comma, search with that and concat the list of addresses before rendering. Removes garbage and seem to expand properly. I'd be interested in how it works for you and if you notice odd behaviour.

PR #500, opened 2022-12-28
Allow to "escape" the confirmation dialog

romanofski:
When we want to abort the composition of an email a confirmation dialog is shown. This allows to either discard or save the mail as a draft.

The only problem is, if we by mistake pressed the keys aborting the composition of the mail and the conformation dialog pops up. Currently there is no way back, you either discard or save the mail as a draft.

That is annoying. Allow aborting the confirmation dialog so you can go back to composing the mail if you wish.


frasertweedale:
LGTM.

PR #502, opened 2022-12-29
fix flake when removing xapian directory

romanofski:
On CI we ran into an error during `tearDown`:

```
file browser handles invalid path input:                  FAIL
    Exception: /tmp/purebredtest-a77e114bdea2381d/Maildir/.notmuch/xapian: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:removeDirectory: unsatisfied constraints (Directory not empty)
    Use -p '/file browser handles invalid path input/' to rerun this test only.
```

The function `removeDirectoryRecursive` will not delete files marked as read only. However the `removePathForcibly` will change read only files in order to force remove the directory.

### DIFF
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -145,7 +145,7 @@ library
                      , notmuch >= 0.3.1 && < 0.4
                      , text
                      , typed-process >= 0.2.8.0
-                     , directory >= 1.2.5.0
+                     , directory >= 1.2.7.0
                      , bytestring
                      , time >= 1.8
                      , case-insensitive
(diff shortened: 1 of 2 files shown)


frasertweedale:
According to https://hackage.haskell.org/package/directory-1.3.8.0/docs/System-Directory.html#v:removePathForcibly , `removePathForcibly` was introduced in *directory-1.2.7.0*.  Can you please bump the lower bound in the cabal file?

frasertweedale (changes requested):
LGTM apart from bumping lower bound.

PR #509, opened 2023-05-21
nix: nix run should work out of box

teto:
`nix run github:purebred-mua/purebred` wouldn't work. One would need to list the flae packages to run ``nix run github:purebred-mua/purebred#purebred` which makes usage harder.

### DIFF
--- a/.nix/overlays.nix
+++ b/.nix/overlays.nix
@@ -32,6 +32,8 @@ let
       '';
       preferLocalBuild = true;
       allowSubstitutes = false;
+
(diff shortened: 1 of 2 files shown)


romanofski:
@teto ooh thanks for the contribution! Is it possible that your clone is a bit behind hence the conflicts?

teto:
indeed. Fixed. I've also made a change that might break anyone using the flake's "overlay" but I doubt that the case and if it is, just replace "overlay" by "overlays.default" it's the "new" (~ a year maybe) format for flake.

PR #516, opened 2024-12-31
upgrade nix dependencies

romanofski:
This updates the flake and since we haven't hacked on this for ages the overlays aren't needed any more since all of them are now available as builds.

Also bumps GHC to be used as any latest GHC 9.2 release (9.2.8 at the time of writing).


romanofski:
uh.. Haskell CI b0rked.. will have to fix that too

frasertweedale:
@romanofski *purebred-email* does not depend on *data-clist* (it isn't in the build plan at all), but *brick* does.  The *data-clist* test-suite component demands `QuickCheck >=2.4 && <2.15`.   I don't know enough about Nix to suggest how to work around this, but a PR to *data-clist* and (if no code changes required) a request to maintainer or Hackage trustees for a metadata revision to relax the constraint upper bound is a sensible step regardless.

PR #517, opened 2025-01-08
fix: allow parsing of mutt aliases using non-ASCII characters

### LINKED ISSUE #515: Loading (mutt) aliases from a file can not handle non-ascii characters
**Describe the bug**
Adding non ASCII characters to the Name fails parsing currently.

The current code uses the wrong charset lookup, that is easily fixed (see `fix/muttaliasparsing`), but configuring it with the icu charsets from `Purebred.Plugin.ICU.icuCharsets` has no bearing on the parsed addresses with non-ASCII characters.

**To Reproduce**
1. Configure an alias file. E.g. I've put mine under `${HOME}/.config/purebred/alias`
2. Add contents (see below)
3. Configure the config file to parse and use the aliases 
6. Compose a new mail
7. For the `To:` use the configured nick and press `<TAB>`

Example alias:
```
alias foo Foo Bah <baz@example.test>
alias bar Füh Bäh <f@example.test>
```

Example purebred config:

```Haskell
main = do
  -- [..]
  addressbook <- initMuttAliasFileAddressBook Purebred.Plugin.ICU.icuCharsets "<path to alias>"
    >>= either (die . show) pure
  --
    purebred
    [ usePlugin . tweakConfig $
          over -- [...]
          . set confAddressBook [addressBook]
          . Purebred.Plugin.ICU.enable
    -- [..]
    ]
```

**Expected behavior**
The email is populated with the configured name in the alias file

**Additional context**
I checked by trying out the parser in a repl, but no luck:
```
λ: import qualified Data.ByteString as B
λ: import Purebred.Plugin.ICU (icuCharsets)
λ: import Data.MIME (address)
λ: import Data.Attoparsec.ByteString (parseOnly)
λ: parseOnly (address icuCharsets) "Foo Bar <foo@example.test>"
Right (Single (Mailbox (Just "Foo Bar") (AddrSpec "foo" (DomainDotAtom ("example" :| ["test"])))))
λ: parseOnly (address icuCharsets) "F\195\188h B\195\164h <f@example.test>"
Left "@: Failed reading: satisfy"
```

There must be something fundamental I'm missing here, since we're already parsing addresses successfully in the application without any problems.

What am I missing?

romanofski:
This uses the improvements from https://github.com/purebred-mua/purebred-email/pull/88 as the current code is unable to parse addresses in the mutt alias file using non-ASCII characters.

This uses the text address parser since the `ByteString` parser would expect the addresses in the mutt alias file to be content encoded. Don't think we want to expect users to do this.

Fixes #515

### DIFF
--- a/configs/aliases
+++ b/configs/aliases
@@ -1,4 +1,4 @@
 alias jen Jennifer Charles <jencharles@example.test>
-alias el "Elwood B.Mack" <elwood@test.example>
+alias el "鈴木 一郎" <elwood@test.example>
 alias bob Robby Cullen <rcullen@example.test>
 alias addy "Addy Mc. Donald" <admcdonald@test.example>
--- a/purebred.cabal
+++ b/purebred.cabal
@@ -149,7 +149,7 @@ library
                      , filepath
                      , mtl >= 2.2 && < 2.4
                      , exceptions
-                     , purebred-email >= 0.5 && <= 0.7
+                     , purebred-email >= 0.7 && <= 0.8
                      , attoparsec
                      , containers
                      , mime-types
--- a/src/Purebred.hs
+++ b/src/Purebred.hs
@@ -129,6 +129,7 @@ module Purebred
   , module Purebred.Plugin,
   module Purebred.Plugin.TweakConfig,
   module Purebred.Storage.Tags,
+  module Purebred.Storage.AddressBook.MuttAliasFile,
   module Purebred.Types,
   module Purebred.Types.Error,
   module Purebred.UI.Actions,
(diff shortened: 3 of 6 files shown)


frasertweedale:
faily test:
```
     exit vim                   (0.03s)
      user error (Text.Regex.Posix.ByteString died: (ReturnCode 8,"Unmatched ( or \\("))
```

romanofski:
Yeah... I'll need to reserve some time to have a look.

romanofski:
Issue is in `tasty-tmux` as it can't match unicode characters. See https://github.com/purebred-mua/tasty-tmux/pull/20

romanofski:
Yay Christmas present... Seems we don't need to jump to a different regex engine alltogether. For whatever reason I don't fully grok yet is that the ByteString with extended UTF8 characters was causing failures. Using Text.encodeUtf8 seems to properly encode the string.

frasertweedale (approved):
ACK.

PR #518, opened 2025-06-22
shut down vty properly

frasertweedale:
Commit e6d247461b0fd5870771256048666a723b0d911b converted to customMainWithDefaultVty, so that we don't have to build the vty ourselves.  But, we still need to shut it down properly.


romanofski:
Can fix the NIX problems after we merge. Up to you. Happy to merge now?

PR #519, opened 2025-06-22
ci: add ghc-9.10 and ghc-9.12 to matrix

frasertweedale:


frasertweedale:
Do not merge til I solve the ghc 9.10 and 9.12 failures :)

romanofski:
That upgrade to Ubuntu 24 is already worth the merge on it's own as the Ubuntu 20 runners are deprecated :)

Issue #520, opened 2025-07-20
b47bd2a breaks `nix build`


lucc:
**Describe the bug**
The commit  b47bd2a breaks `nix build`

**To Reproduce**
1. Run `nix build nix  build --option experimental-features "flakes nix-command" github:purebred-mua/purebred?ref=b47bd2a2a42cb8374adad7f3ee9256e355e5da28`
2. See the build failure

**Expected behavior**
The purebred derivation should be build.

**Additional context**

The parent commit (2a9ffea39014b31cfed890e0fff4c1b8649fa845) can be build just fine.

<details>
<summary>
Build logs for b47bd2a
</summary>

~~~
@nix { "action": "setPhase", "phase": "setupCompilerEnvironmentPhase" }
setupCompilerEnvironmentPhase
Build with /nix/store/xdahirc0rarz4k39qlm8ykyzqs3axfzd-ghc-9.2.4.
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking sources
unpacking source archive /nix/store/c922i1l858d7g3x3bfkz4nn8v2yc4k0z-qljpc8hmilxi5s8m39z0v4qal1b20z94-source
source root is qljpc8hmilxi5s8m39z0v4qal1b20z94-source
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "compileBuildDriverPhase" }
compileBuildDriverPhase
setupCompileFlags: -package-db=/build/tmp.oHJYvNmvOa/setup-package.conf.d -j4 +RTS -A64M -RTS -threaded -rtsopts
[1 of 1] Compiling Main             ( Setup.hs, /build/tmp.oHJYvNmvOa/Main.o )
Linking Setup ...
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
[...]

PR #521, opened 2025-07-20
ci: update cache action

lucc:
The builds are currently failing because the infrastructure behind the old version of the cache action was migrated:
https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

There also is a note on the actions ui in github: https://github.com/purebred-mua/purebred/actions/runs/15842188147/job/44656791801

This might be a reason why #520 was not noticed by ci.

### DIFF
--- a/.github/workflows/nixos-ci-flakes.yml
+++ b/.github/workflows/nixos-ci-flakes.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Local cache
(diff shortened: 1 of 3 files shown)


romanofski:
I'd be happy to merge this for now as `master` is broken worse without it. The 9.10 and 9.12 failures seem to be needing code changes as well.

PR #522, opened 2025-08-19
fix plugins on GHC >= 9.10

frasertweedale:
GHC 9.10 made instance resolution more strict, and no longer derives superclasses in cases where `UndecidableInstances` is in play. Compilation succeeds, but *using* plugins can fail where a plugin has hooks that use fewer capabilities than the power available for that hook (e.g. a `ConfigHook Pure`).

To resolve the issue, explicitly include relevant superclasses in quantified constraints, even where they are seemingly redundant.

Add a test function to ensure (at library compile time) that that we did not miss any important cases.  The test make plugins for all hooks at all capability levels and applies 'usePlugin' to each.


romanofski:
\o/

PR #523, opened 2025-08-25
deps: drop support for GHC < 9.6

frasertweedale:
The platforms we care about all have GHC 9.6 or later, so drop support for older versions.  Clean up redundant extensions and imports, and avoid some warnings.


romanofski:
Happy to adjust the Nix side :)

PR #527, opened 2025-12-19
CI: only run on pull request

romanofski:
Currently we run builds on push *and* pull request. However I find the push run superfluous given that we often just create/iterate on the pull request anyway. No point of running everything twice.

This changes the behaviour to run on:

* master merge
* pull requests when they're opened or edited.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -14,8 +14,13 @@
 #
 name: Haskell-CI
 on:
-  - push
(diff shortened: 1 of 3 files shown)


frasertweedale:
nack, I prefer running them on all pushes on all branches.  A branch doesn't *necessarily* have an associated PR.

Not like it's costing us anything...

romanofski:
Ok. Not a big deal.

PR #528, opened 2025-12-25
feat: allow to reply from a specific entity

romanofski:
This is one of those odd ones. I prefer to view text/plain, however when I received mails from an insurance company it had multipart mixed entities while the actual mail was in the first text/html entity and the additional parts essentially contained some disclaimer.

It's currently possible to view all entities (even tho if you prefer text/plain it picks unfortunately the wrong entity), but reply is difficult because Purebred picks the text plain part which is unfortunately breaking conversation.

So this is a way to be able to reply to a specific entity rather than your preferred pick.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1757981991,
-        "narHash": "sha256-vR6xwDIpFHAd+YhJaVK58mVsbgXEUm+8ORCEsOqy5lM=",
+        "lastModified": 1766622272,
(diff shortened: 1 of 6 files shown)


frasertweedale:
ACK from me.  Merge at will.

Relatedly, I was thinking about adding a function or plugin to configure/register a per-sender preferred representation, so you can override the default entity picker heuristic for particular senders.  Use case is exactly what you describe - multipart where e.g. text/plain part is just "look at the HTML" or similar stupidity.

romanofski:
Perhaps we should look at moving more functionality to plugins?

Issue #9, opened 2017-07-18
History of searches


romanofski:
A list of recently performed search patterns could be very useful. This can be transformed into a little editor and/or if maybe sorted on recently used serves as an easy way to perform common searches without too much fuzz...

frasertweedale:
A way to save searches too.  But IMO this is not a baseline feature.

(This gets me thinking... should we create a milestone for what we we consider a critical feature and what we can kick down the road?  Or issue labels?)

romanofski:
Related to #280. If we are able to Haskeline, we would get the history for free.

PR #529, opened 2026-06-04
Haskeline support in Purebred

romanofski:
This has still a few rough edges and I need to test it a bit more. 

I've used two widgets I converted:
1. tag editing
2. opening commands

I think if we want to go ahead it will probably make sense to move Brick/Haskeline into it's own package rather than keeping it in Purebred.


romanofski:
@frasertweedale if you want to take a peek? Not ready to merge yet. Just thought I post it in case there a few things you think should be different.

romanofski:
This should fix #9

